PR #68, opened 2015-02-26
allow omitting of delta values for moveHandlers

frederik-b:
Allow to move Shapes and Connections in only one direction by omitting the x or y -delta (or both to make a dummy move).
My next Pull request will build upon this feature, but could be easily adjusted to work without this feature.


frederik-b:
omitting delta values messes up the commandStack.undo(); method

nikku:
I will close this PR. We are working on an alternative implementation for this.

PR #69, opened 2015-02-27
fix(overlays): prevent display of overlays outside parent container

SebastianStamm:
Without overflow:hidden in the container element, the html overlay elements are rendered outside the parent container
bpmnioclipping

### DIFF
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -37,6 +37,7 @@ function createContainer(options) {
 
   assign(parent.style, {
     position: 'relative',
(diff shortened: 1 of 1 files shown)


nikku:
I think it is totally fine to have this kind of behavior (i.e. display overlays outside the parent).

You may easily customize the behavior yourself though by including a css declaration like the following:

```
.djs-container {
  overflow: hidden;
}
```

nikku:
Merged via 6bb7caeaeb235125324e3b84b99559c0711f1d09

PR #70, opened 2015-02-27
feat(markers): add has and toggle marker methods

SebastianStamm:


nikku:
I added a bunch of tests for the exposed marker methods.

Thanks for the contribution.

Closed via 194246b0fa572b00bf83d79cad4995a881252a0e

PR #72, opened 2015-03-02
impr(zoomscroll): normalize zoom factor

SebastianStamm:
This PR improves the zoomscroll functionality by normalizing the zoom factor. When the user zooms out and then zooms in again, the zoom level is the same as before. With the previous solution, the diagram would be slightly zoomed out.


ISO50:
Thx for this patch :smiley:

PR #74, opened 2015-03-13
feat(core/Canvas): allow centering of diagram with fit-viewport

SebastianStamm:


nikku:
Could you add some test cases that verify your code is actually working?

nikku:
Merged via 0a9265a3df1c0d2776c32f87e51d368cb0b966e8

PR #90, opened 2015-06-12
chore(command): set dependency to EventBus only

SebastianStamm:
The command module requires the whole core. This includes diagram-js specific things like Canvas and GraphicsFactory.

In order to make the Command and EventBus module independent from diagram-js, this PR changes the dependency of the command module to the EventBus only.


nikku:
Merged via f13de340e5b46e6f1a9e7783e7d9b330282fdb2c

PR #115, opened 2015-09-12
Symbols caused problems with Browserify

phalix:
Exchange a symbol " " with an actual space. The output files of browserify caused problems because they could not handle the symbol.


ricardomatias:
@phalix, which OS are you using ? I don't have any problems with browserify under OS X.

nikku:
The problem is that these are special _non breakable_ white-space characters Mac users tend to accidentally add via `SHIFT+SPACE` or the like.

I had the same issues on Windows / Linux with older versions of browserify at least.

We should probably just merge this.

PR #127, opened 2015-11-13
fix(overlays): respect min/max zoom level when overlay is added

### LINKED ISSUE #126: Zoom levels are not respected when overlay is added
Reproduce

1. Click on a element to open context pad (context pad A)
2. Zoom out until we're below the min zoom level and overlay (the context pad) disappears (Expected)
4. Now click on another element to open a (new) context pad (context pad B)
5. Context pad B is shown, even though we are below the min zoom level (**Not Expected**)
6. Zoom out once more (still below min zoom level) to trigger a update
7. Context pad B is now hidden (Expected)

Conclusion

The zoom level is not respected when a overlay is added, only when the canvas is changed.
When below the min zoom-level, clicking on elements that have been clicked BEFORE we zoomed will not display their context pads, while clicking on elements that have never been clicked on before will open their context pads.

adbre:
Closes #126

### DIFF
--- a/lib/features/overlays/Overlays.js
+++ b/lib/features/overlays/Overlays.js
@@ -417,23 +417,30 @@ Overlays.prototype._addOverlay = function(overlay) {
   this._overlays[id] = overlay;
 
   this._updateOverlay(overlay);
+  this._updateOverlayVisibilty(overlay, this._canvas.viewbox());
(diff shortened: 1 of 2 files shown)


nikku:
Thanks for the fix!

PR #129, opened 2015-11-30
chore(test): Add .defined matcher

adbre:
Some tests (spec/features/modeling/DeleteElementsSpec.js is one example)
expects some value .to.be.defined or .to.not.be.defined.
But Chai does not have a .defined assertion. It's called .undefined.

Since .defined is undefined, the assertion is a no-op and won't detect
broken code.

I have tested this by commenting out the WHEN statement in
DeleteElementsSpec.js. Without this fix, the test will then silently
continue. With this fix the error is detected as expected.

### DIFF
--- a/test/TestHelper.js
+++ b/test/TestHelper.js
@@ -12,3 +12,4 @@ TestHelper.insertCSS('diagram-js-testing.css',
 // add suite specific matchers
 global.chai.use(require('./matchers/BoundsMatchers'));
 global.chai.use(require('./matchers/ConnectionMatchers'));
+global.chai.use(require('./matchers/DefinedMatchers'));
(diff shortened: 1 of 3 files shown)


nikku:
Good catch! Around 100 matches in diagram-js...

That looks like quite a left over from our previous Jasmin TDD setup. :stuck_out_tongue_winking_eye: 

I am wondering if it may be a better solution to migrate the project (and bpmn-js) to the`to(.not).exists` matcher that comes built in with Chai.

nikku:
Just checked it. Looks like we can safely use `(.not).exists` instead.

I'll push a fix for this soon. Thanks for having a close look on our test suite!

nikku:
Closing this in favour of bpmn-io/bpmn-js#423

PR #138, opened 2016-01-26
Zoom scroll fix 416

janstuemmel:
Issue: https://github.com/bpmn-io/bpmn-js/issues/416

You can now disable zoomScroll via config:

```javascript
var renderer = new BpmnModeler({ container: canvas, zoomScroll: { disabled: true } });
```

and enable it programmatically:

```javascript
renderer.invoke(function(zoomScroll){
   console.log("Invoked ZoomScroll")
   zoomScroll.setEnabled(true);
});
```

what does not work:
if zoom scroll is enabled i cant disable it via invoke. unbinding the domEvent does not work.


nikku:
I squashed and rebased your changes and merged them with some adjustments (1fe8f099d9daae5823a07f0896cbba5d9a873025).

PR #141, opened 2016-02-18
Declare `modeling` dependency on used by Move.js

jeromew:
I am still trying to understand the project but it seems that this dependency is missing in the declarations as `modeling` is used inside Move.js

I hope this helps. Congratulations on your diagram framework. It looks very impressive even if very intimidating !

### DIFF
--- a/lib/features/move/index.js
+++ b/lib/features/move/index.js
@@ -1,6 +1,7 @@
 module.exports = {
   __depends__: [
     require('../interaction-events'),
+    require('../modeling'),
     require('../selection'),
     require('../outline'),
     require('../rules'),


nikku:
`modeling` is an implicit dependency that you'd need to provide as part of your modeling solution. The reason is that it is highly specific to your solution.

```javascript
var diagram = new Diagram({
   modules: [
     require('diagram-js/lib/features/move'),
     require('my-custom-modeling-module')
  ]
});
```

So what you'd need to do is providing `modeling` via `my-custom-modeling-module` i.e. by extending `diagram-js/lib/features/modeling`.

PR #154, opened 2016-04-01
fix(CommandStack): Given array, dirty elements should exclude undefined

adbre:
This commit will prevent undefined elements to being concated to
CommandStack's dirty collection.

If the command handler returned a single item and that item would be
a undefined reference, CommandStack will prevent it from being tracked
as dirty.

However, if a command handler returns an array containing a undefined
element, the array will simply be concated to the dirty collection
without filter the undefined element.

The above can be reproduced in the current version of bpmn-js (v0.14.1)

  1. Create a gateway
  2. Create a second element (task for example)
  3. Create a sequence flow connection from the gateway to the task
  4. Make the sequence flow default
  5. Remove the sequence flow connection
  6. Create a sequence flow connection from the gateway to the task (new)
  7. Make the sequence flow default

The 7th step will produce an exception as the UpdatePropertiesHandler
will return undefined as the first sequence flow no longer exists, but
the gateway's default property was not cleared in step 5.

### DIFF
--- a/lib/command/CommandStack.js
+++ b/lib/command/CommandStack.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var unique = require('lodash/array/unique'),
+    filter = require('lodash/collection/filter'),
     isArray = require('lodash/lang/isArray'),
     assign = require('lodash/object/assign');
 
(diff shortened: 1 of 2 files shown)


nikku:
We fixed this in bpmn-io/bpmn-js#505 and decided to not merge this PR to disguise future mistakes.

adbre:
I thought of that as well, but shouldn't the `if(elements)  return;` be removed then as well? Seems odd to have protection for one scenario and not the other. I only tried to make it conistent.

adbre:
`if(!elements)` of course

nikku:
What line are you referring to?

adbre:
https://github.com/bpmn-io/diagram-js/blob/master/lib/command/CommandStack.js#L431

nikku:
Well, a handler may decide to not mark any element as dirty.

This is legacy but cannot easily be changed right now (if I remember correctly the delete handlers are not properly returning elements).

adbre:
OK

PR #179, opened 2016-06-20
fix gh issue 570

TimSusa:
Get rid of dropping last Hyphen in the remaining line. Grunt auto-test ran without any issues, afterwards.


ricardomatias:
Could you please add a test case for this ?

TimSusa:
Ricardo, 

sry for the (holiday-) delay and thanks for the hint. I have added a simple test, that extracts the text content from the created element and checks that string to have a hyphen as last char.

nikku:
Closed via 7baa0265e6760fcfe74432873a75e7d9ea80d0c1. Thanks for the PR.

PR #182, opened 2016-07-05
fix(modeling): keep parent of connection on primary selection with different parent

ulasakdeniz:
Closes bpmn-io/bpmn-js#525


nikku:
Merged via https://github.com/bpmn-io/diagram-js/commit/acca915f45912bdb1c8e3cf3672b830aa3c7504f. Thanks for the PR!

PR #185, opened 2016-07-27
feat(modeling): Model can be read-only

adbre:
This commit will add support for disabling modeling actions, but will leave navigation tools enabled.

To get or set the read-only state, the modeling module exposes a readOnly function.

    modeling.readOnly(true); // enable read-only
    modeling.readOnly(false); // disable read-only
    modeling.readOnly(); // will leave current state unchanged, returning true or false

The requirements I have for a read-only state are
- Modeling actions should not be permitted
- Disabled actions should not be displayed (palette and context-pad)
- Navigation (pan and zoom) should be permitted
- Selection should still be possible even in a read-only state
- Toggling read-only state should be possible programatically
- When read-only state is activated, any ongoing modeling user actions should be canceled

I am expecting comments and I will update the PR to align with any feedback given, small or big.

A live demo is available at https://adbre.github.io/bpmn-js-readonly-example/ (use the lock icon in the upper-right corner to toggle the read-only state).
Source code for the demo can be obtained at [adbre/bpmn-js-examples/disable-modeling/modeler](https://github.com/adbre/bpmn-js-examples/tree/disable-modeling/modeler).

This PR addresses the [use case I posted in the forums](https://forum.bpmn.io/t/changing-from-modeler-to-viewer-and-back-again/900/1) recently.

Please note, this PR also requires changes in bpmn-js, see bpmn-io/bpmn-js#598

### DIFF
--- a/lib/features/context-pad/ContextPad.js
+++ b/lib/features/context-pad/ContextPad.js
@@ -44,8 +44,15 @@ ContextPad.prototype._init = function() {
 
   var self = this;
 
-  eventBus.on('selection.changed', function(e) {
+  eventBus.on('readOnly.changed', function (e) {
+    self._readOnly = e.readOnly;
 
+    if (e.readOnly && self.isOpen()) {
+      self.close();
+    }
+  });
+
+  eventBus.on('selection.changed', function(e) {
     var selection = e.newSelection;
 
     if (selection.length === 1) {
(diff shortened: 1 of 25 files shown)


adbre:
Abandoning this PR, as a different approach is preferred.
See the [forum thread](https://forum.bpmn.io/t/changing-from-modeler-to-viewer-and-back-again/900/7?u=adbre) for info.

PR #193, opened 2016-08-31
fix(assets/diagram-js.css) Removes semicolon after the curly bracket.

erdemkeren:


pedesen:
Thanks for the contribution. Merged via https://github.com/bpmn-io/diagram-js/commit/65856846eacbdd3b7833392ecefdf2e011b8d296.

PR #196, opened 2016-10-26
Support array of classNames when adding entries

narve:
feat(palette): When adding entries, allow className to be an array of strings.

### DIFF
--- a/lib/features/palette/Palette.js
+++ b/lib/features/palette/Palette.js
@@ -168,8 +168,8 @@ Palette.prototype._update = function() {
       }
 
       if (entry.className) {
-        const classNames = Array.isArray(entry.className ) ? entry.className : [ entry.className ];
-        classNames.forEach(e => domClasses(control).add(e));
+        var classNames = Array.isArray(entry.className ) ? entry.className : [ entry.className ];
+        classNames.forEach( function(e) { domClasses(control).add(e); });
       }
 
       if (entry.imageUrl) {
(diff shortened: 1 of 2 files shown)


narve:
Sorry, technical error - updated the PR with the code that actually passed the tests :)

nikku:
In order to merge this PR we need a test case verifying the newly added functionality.

narve:
Can I interpret this as "you generally agree to the requirement/functionality/solution"? If so I can add a test case and update the PR. 

I also considered allowing className to contain spaces and split the string, but in general I prefer a more structured approach.

nikku:
> Can I interpret this as "you generally agree to the requirement/functionality/solution"? If so I can add a test case and update the PR.

Yes, makes sense all together. Maybe we can go for both, space separated class list and array?

narve:
Hm... I prefer the array-approach... then I don't have to research what "space" actually, is according to html-specs, and how to split on that in javascript :)

As long as it is possible to use an array, and this is a plausible (obviously, since I needed it) but not very common requirement, I believe that arrays should be good enough.

nikku:
Merged with some changes via https://github.com/bpmn-io/diagram-js/commit/59438339a8ad29bac790804bc56497447b63b0b1.

Added the same functionality for the context pad, too: https://github.com/bpmn-io/diagram-js/commit/53cc1d431da2b9a112fa8329ca4eff3a45e586be

nikku:
Thanks for your contribution.

narve:
Superb, thanks! I was just writing a comment, asking whether I should update the PR or not, when the page updated itself  :)

PR #203, opened 2017-01-05
add multi label support for base model element, adjust test 

KosrenDQ:
Add multi label support to the base element, therefore it can be used in connections and shapes.

### DIFF
--- a/lib/model/index.js
+++ b/lib/model/index.js
@@ -6,7 +6,7 @@ var assign = require('lodash/object/assign'),
 var Refs = require('object-refs');
 
 var parentRefs = new Refs({ name: 'children', enumerable: true, collection: true }, { name: 'parent' }),
-    labelRefs = new Refs({ name: 'label', enumerable: true }, { name: 'labelTarget' }),
+    labelRefs = new Refs({ name: 'labels', enumerable: true, collection: true }, { name: 'labelTarget' }),
     attacherRefs = new Refs({ name: 'attachers', collection: true }, { name: 'host' }),
     outgoingRefs = new Refs({ name: 'outgoing', collection: true }, { name: 'source' }),
     incomingRefs = new Refs({ name: 'incoming', collection: true }, { name: 'target' });
@@ -47,10 +47,12 @@ function Base() {
   parentRefs.bind(this, 'parent');
 
   /**
-   * @name Base#label
+   * The list of labels
+   * 
+   * @name Base#labels
    * @type Label
    */
-  labelRefs.bind(this, 'label');
+  labelRefs.bind(this, 'labels');
 
   /**
    * The list of outgoing connections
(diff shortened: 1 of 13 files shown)


nikku:
That was a quick PR, thanks for your contribution!

I got one change request: Can we still provide the `element.label` property as the primary label? This would allow existing toolkits that rely on the single label to continue working. Assigning the label would store it in `element.labels[0]` and reading it would simply read it from `element.labels[0]`. 

Something on these lines should be achievable using `Object.defineProperty(...)`.

KosrenDQ:
I readded the label property and revert the test to the single label usage.

nikku:
Thanks for the changes. I'll try to have another look on your PR this week.

lib/model/index.js
```diff
@@ -39,6 +39,17 @@ function Base() {
   });
 
   /**
+   * Single label support, will mapped to multi label array
+   * 
+   * @name Base#label
+   * @type Object
+   */
+  Object.defineProperty(this, 'label', {
+    get: function() { if (this['labels'][0]) { return this['labels'][0]; } else { return null; }}, //Ensure that the getter returns a label or null
```
nikku: The reason you need to currently return `null` rather than `undefined` is that parts of the model wiring (especially setting `someElement.labelTarget = label`) does not work with the changes you provide.

I'm currently investigating, if there is an easy fix.

nikku: Does not look like something we can easily fix. Investigating...

nikku:
Thanks again for the PR. It looks fine overall. I added a few minor changes (see above).

What is missing is test cases that verify that the toolkit properly handles multiple labels during 

* [ ] element move (move labels with element)
* [ ] attached label move (moves labels, too)
* [ ] element deletion (remove labels with it)
* [ ] copy paste

I added a bunch of test cases for the model and refined a few deletion test cases, as these are the places where it matters most to us. 

As we'll not use multi-label support any time soon, it makes sense to add test cases for the aspects listed above, too so that we do not break or unintentionally remove these features in the future.

nikku:
Any plan to follow up on my review feedback?

KosrenDQ:
Yeah, i want to implement the test cases you describe. Unfortunately the last week was full of stress. Next week will be a good one for science.

KosrenDQ:
So I added a couple of tests, I hope it covers all your cases.

nikku:
Thanks, I'll look into the additions in the next days.

nikku:
I squashed your changes and did a little bit of cleanup before I realized the PR still misses test cases for copy and paste.

Trying to add them I realized pasting of labels is currently broken, as it was before: https://github.com/bpmn-io/diagram-js/issues/206. 

We'll have a look inside and see that we get the stuff fixed before merging the PR.

nikku:
Merged at last via https://github.com/bpmn-io/diagram-js/commit/1b5fa55268137940ee7864294d255d309282677c.

PR #228, opened 2017-09-21
Add ability to create non-scaling overlays

vpanfilov:
I propose to add ability to create non-scaling overlays, which retain the same size regardless of current diagram zoom. So I added boolean `scale` attribute to overlay configuration properties.

For now, all overlays scales and moves with the diagram, because overlay container is transformed on zoom and scroll events. How can we allow some overlays to be scale-independent in this situation? The idea is simple: we should revert scale for individual overlays using `transform: scale(1/zoom, 1/zoom)` and `transform-origin: top left` CSS properties.

The same feature can also be easily added to **tooltips** module.

### DIFF
--- a/lib/features/overlays/Overlays.js
+++ b/lib/features/overlays/Overlays.js
@@ -96,7 +96,8 @@ function Overlays(eventBus, canvas, elementRegistry) {
     show: {
       minZoom: 0.7,
       maxZoom: 5.0
-    }
+    },
+    scale: true
   };
 
   /**
@@ -191,6 +192,7 @@ Overlays.prototype.get = function(search) {
  * @param {Number}                  [overlay.position.top]       relative to element bbox top attachment
  * @param {Number}                  [overlay.position.bottom]    relative to element bbox bottom attachment
  * @param {Number}                  [overlay.position.right]     relative to element bbox right attachment
+ * @param {Boolean}                 [overlay.scale]              false to preserve the same size regardless of diagram zoom  
  *
  * @return {String}                 id that may be used to reference the overlay for update or removal
  */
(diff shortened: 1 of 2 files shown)


nikku:
Thanks for your pull request!

I like the feature you've implemented and got one more thing that may be worth discussing: For non-scaling overlays, how much sense does it make to hide it on small zoom levels?

I'd assume that it is always visible, as it is always in an appropriate size.

vpanfilov:
As I understand, we can easily create separate defaults for non-scaling overlays and apply them in `Overlays.prototype.add` function. But who will determine the best default values?

nikku:
We discussed this internally. 

Right now, what is on top of our heads is:

* Change the default for overlays to always show them; still keep the ability for users to hide overlays via `show.minZoom`, `show.maxZoom`
* Add the option for users to decide whether to scale, not to scale or scale up / down overlays
* Take actual overlay scaling into account when determining whether the overlay should be shown or not
  (this will make sure that overlays with `scale` false will effectively never be hidden, as they are always be shown at `scale = 1`)

nikku:
Work in progress interaction between overlays:

overlays

vpanfilov:
Your first statement seems most reasonable for me. If user did not provide `minZoom` and `maxZoom` overlay should **always** be visible. It will be a breaking change, so do not forget to double mention it in release notes if you decide to go this way.

> Add the option for users to decide whether to scale, not to scale or scale up / down overlays

I can't imagine any use case for up-scaling right now. Maybe it will be more optimal to let user provide either **scale:** _true/false_ or some scaling function **f(zoom)**?

> Take actual overlay scaling into account when determining whether the overlay should be shown or not

Don't really understood what do you want to achieve by that. In my opinion, `minZoom` and `maxZoom` options are enough.

nikku:
Thanks for your feedback. We'll figure something out.

nikku:
We'll merge your changes as part of https://github.com/bpmn-io/diagram-js/pull/230. Thanks for the PR.

PR #244, opened 2018-01-16
feat(interactionEvents): add contextmenu event

bort2012:
* add new event "contextmenu" to InteractionEvents
* add new the new event to the current spec of InteractionEvents

BREAKING CHANGE:
remove "isPrimaryButton" import and remove the if block in the fire method. This is needed to not ignore right clicks anymore.

before:
```
// only react on left mouse button interactions
// for interaction events
if (!isPrimaryButton(event)) {
  return;
}
```

after:
block is removed

Closes bpmn-io/bpmn-js#750

### DIFF
--- a/lib/features/interaction-events/InteractionEvents.js
+++ b/lib/features/interaction-events/InteractionEvents.js
@@ -3,6 +3,8 @@
 var forEach = require('lodash/collection/forEach'),
     domDelegate = require('min-dom/lib/delegate');
 
+var isPrimaryButton = require('../../util/Mouse').isPrimaryButton;
+
 var svgAppend = require('tiny-svg/lib/append'),
     svgAttr = require('tiny-svg/lib/attr'),
     svgCreate = require('tiny-svg/lib/create');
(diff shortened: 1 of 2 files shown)


nikku:
Thanks for your PR. 

Unfortunately removing the `isPrimaryButton` constraint will break pretty much all existing applications once the user right clicks on the diagram.

#### Old behavior

* right clicks are ignored

#### New behavior

* right clicks trigger diagram interactions

Please think of a solution that implements the `contextmenu` event without removing that constraint for the other event types (such as `element.click`).

bort2012:
Thank you for your feedback.

I pushed a new solution without removing the "isPrimaryButton" constraint. 

About the solution:
First I thought about adding a second bindings-object to store the "right-click-enabled" bindings like "contextmenu" and register them to the mousehandler with a Flag on the event that its enabled for right clicks. 

But I ended up with a solution that stores all bindings in the existing object (like current behavior) and just saves the "right-click-enabled" bindings in an array. Now everytime an interactionEvent gets fired it checks the constraint "isPrimaryButton" and if it is in the "right-click-enabled" array. 
I chose this solution to not touch the current logic.

nikku:
I've combined your changes into a single commit and merged everything via https://github.com/bpmn-io/diagram-js/commit/4f3df864dd4d506a6f37b0ffa49217b03169a306.

nikku:
Thanks.

Issue #234, opened 2017-10-23
Migrate to min-dash


nikku:
Right now we use [lodash](https://lodash.com/) as a utility library. Unfortunately, it has a massive footprint.

To give some stats right now the lodash helpers used in [bpmn-js](https://github.com/bpmn-io/bpmn-js) make up for `56K` (BPMN viewer) and `90K` (BPMN modeler) of the overall bundle size (unminified).

Even when minified, that amout of _utility_ is significant.

We can do better. [min-dash](https://github.com/bpmn-io/min-dash) provides us with all the utilities we need while being only `14K` in size and 4K minified. Plus, min-dash properly tree-shakes when using a current, ES2016 aware bundler (i.e. `webpack > 2`). 

__Tasks__

Migrate the following projects to min-dash:

* [x] bpmn-js
* [x] diagram-js
* [x] [bpmn-moddle](https://github.com/bpmn-io/bpmn-moddle/tree/min-dash)
* [x] [moddle](https://github.com/bpmn-io/moddle/tree/min-dash)
* [x] [moddle-xml](https://github.com/bpmn-io/moddle-xml/tree/min-dash)

TheSharpieOne:
What is the remaining lift for this? I see progress seems to have stopped halfway through.
Just diagram-js and bpmn-js is left. Does min-dash contain all of the utilities needed for the migration or is there work to be identified and done on that front?

nikku:
> I see progress seems to have stopped halfway through.

Well, bpmn-js and diagram-js are most of the work :wink:.

> Does min-dash contain all of the utilities needed for the migration or is there work to be identified and done on that front?

Most of the time there exists simple drop-in replacements in min-dash for the lodash utlities.

Three cases need special handling:

* similar to lodash@4 min-dash does not support the context (or _this_) parameter
* min-dash only contains `sortBy` and `uniqueBy` and not the more general unique properties we use in some places in the code base
* we use some obscure lodash functions in some places that we need to get rid of / replace by with simpler variants

TheSharpieOne:
FWIW (and to prevent duplicate efforts), I'll take this on.

nikku:
Great to hear you're giving it a try.

Please use the following style as a general pattern for requiring the utilities: 

```javascript
var forEach = require('min-dash').forEach,
    sortBy = require('min-dash').sortBy;
```

This makes later migration to ES6 easier and allows us to tree-shake (i.e. only include the utilities we actually need) during bundling.

[...]

PR #246, opened 2018-03-15
[WIP]  chore(project): upgrade lodash to min-dash

TheSharpieOne:
Fixes https://github.com/bpmn-io/diagram-js/issues/234

### Proposed Changes

* Switch out lodash for min-dash
* Add some utils for functions which lodash had and min-dash will not provide (fill the gap)

### DIFF
--- a/lib/command/CommandStack.js
+++ b/lib/command/CommandStack.js
@@ -456,7 +456,7 @@ CommandStack.prototype._popAction = function() {
   actions.pop();
 
   if (!actions.length) {
-    this._eventBus.fire('elements.changed', { elements: uniqueBy(function (e) { return e; }, dirty) });
+    this._eventBus.fire('elements.changed', { elements: uniqueBy(function(e) { return e; }, dirty) });
 
     dirty.length = 0;
 
--- a/lib/command/CommandStack.js
+++ b/lib/command/CommandStack.js
@@ -1,8 +1,8 @@
 'use strict';
 
-var unique = require('lodash/array/unique'),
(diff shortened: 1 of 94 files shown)


lib/command/CommandStack.js
```diff
@@ -456,7 +456,7 @@ CommandStack.prototype._popAction = function() {
   actions.pop();
 
   if (!actions.length) {
-    this._eventBus.fire('elements.changed', { elements: unique(dirty) });
+    this._eventBus.fire('elements.changed', { elements: uniqueBy(function(e) { return e; }, dirty) });
```
TheSharpieOne: min-dash does not support comparing objects without a given key/property or transforming the object into a key (string or number). I am not sure what key/property would be best to determine if the objects in dirty are unique.
Currently it just checks if the objects are the same object, but with they are both different objects with the same values, there is no way for min-dash's unique to determine that.
I could stringify the object and which would allow for comparing the string, but that is performant and I believe different property orders will result in different strings even if all of the properties are the same between the objects.

nikku: The elements `id` is the key you're searching for.

nikku:
I've applied a few changes before merging your changes via https://github.com/bpmn-io/diagram-js/commit/b5f17f786a16f8cb68ecd4118dcd05d17eda85b6:

* using `matchPattern({ id: foo })` with find instead of `find(a, function(...) { })`
* using `flatten`, `size`, `has`, `merge` utilities (newly implemented in min-dash)

PR #250, opened 2018-04-18
Use delta function instead of subtract

nikeee:
This PR removes the duplicate function `Math#substract` and uses the more often used function `PositionUtil#delta`.


nikku:
For compatibility reasons we cannot simply re-move `Math#substract`. Could you amend your PR to re-export `PositionUtil#delta` as `Math#substract`?

nikeee:
Done

nikku:
Thanks!

PR #254, opened 2018-05-09
chore(popup-menu): allow multiple popupMenuProvider

### LINKED ISSUE #253: Creating an additional PopupMenuProvider should not overwrite ReplaceMenu entries
### Expected Behavior

I would expect it to be possible to register an additional PopupMenuProvider. The registered popup menu should open on click on a custom context pad entry. I would expect that this does not interfere with the replace menu (which is also a PopupMenuProvider).

### Actual Behavior

On ContextPad open:
* the ReplaceMenu entries get created first and stored in `_current.entries` and `_current.headerEntries`
* the additional PopupMenu entries are created second, overwriting `_current.entries` and `_current.headerEntries`

On ReplaceMenu open:
*  the PopupMenu opens with the (wrong) entries of the additional PopupMenu

### Steps to reproduce the Behavior

* create a PopupMenuProvider and register it in PopupMenu
* click on element
* open context pad
* open replace menu 
* the replace menu opens with the wrong content

### Proposed solution

The lifecycle of the `_current` object in `PopupMenu` needs to change in order to achieve this. Ideally the current popup menu entries are created when opening the popup menu, and not already when opening the context pad. It probably makes sense to merge `PopupMenu.create` into`PopupMenu.open` and then only call it on a click on the respective context pad entry. This would imply a breaking change in `PopupMenu`.

pedesen:
Prevoiusly the popupMenu#create needed to be called when
context pad was opening. This is no longer needed. Now
only popupMenu#open needs to be called on context pad entry
click.

BREAKING CHANGES:

* popupMenu#create was removed
* popupMenu#open now expects element, id, position as params
* popupMenu#isEmpty now expects element, providerId as params

Fixes #253

### DIFF
--- a/lib/features/popup-menu/PopupMenu.js
+++ b/lib/features/popup-menu/PopupMenu.js
@@ -30,6 +30,7 @@ export default function PopupMenu(eventBus, canvas) {
   this._eventBus = eventBus;
   this._canvas = canvas;
   this._providers = {};
+  this._current = {};
 }
 
 PopupMenu.$inject = [ 'eventBus', 'canvas' ];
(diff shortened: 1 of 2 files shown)


nikku:
The current implementation ensures that only one PopupMenu can be open at a time.

Will this be enforced under the new behavior, too?

pedesen:
Yes, a popup menu which is currently open for a specific element will be closed before opening another one. https://github.com/bpmn-io/diagram-js/pull/254/commits/b1852e1d71f67bd36ae1eb02748d2d0cbf124625#diff-33a9c650347c9b77ab2fd2d3e67dee12R114

PR #255, opened 2018-05-14
feat(popup-menu): support multiple classnames on popup-menu entries

felixlinker:
### Proposed Changes

* A popup-menu entry can now accept multiple classes as `className` each of them delimited by a space, e. g. `className: 'class1 class2'`

### DIFF
--- a/lib/features/popup-menu/PopupMenu.js
+++ b/lib/features/popup-menu/PopupMenu.js
@@ -375,7 +375,7 @@ PopupMenu.prototype._createEntry = function(entry) {
   entryClasses.add('entry');
 
   if (entry.className) {
-    entryClasses.add(entry.className);
+    entry.className.split(' ').forEach(className => entryClasses.add(className));
   }
 
   domAttr(entryContainer, DATA_REF, entry.id);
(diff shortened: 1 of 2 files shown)


nikku:
What is the use-case for this?

felixlinker:
I'm using the modeler in an "embedded" scenario. In my case, it is used in Wordpress. I'd like to add a popup-menu entry that has one of the pre-existing icons that come shipped with wordpress displayed. To use one of those, I need to add two classes: `dashicons-before dashicons-ICON_NAME`.

However, this scenario applies to many frameworks, such as bootstrap. There you also need to add two classes to a `<span>` in order for it to display a glyphicon. ([Click here for an example](https://getbootstrap.com/docs/3.3/components/))

nikku:
Sounds legitimate!

Could you add test coverage that ensures your contribution works? [`PopupMenuSpec`](https://github.com/bpmn-io/diagram-js/blob/master/test/spec/features/popup-menu/PopupMenuSpec.js) is the right place to do so.

felixlinker:
Done ✔️

nikku:
Thanks.

PR #256, opened 2018-05-28
feat(util/Text): make text-layouting lineHeight aware

nikku:
This ensures we properly align text and support a numeric `lineHeight` property.


nikku:
Need to improve external label rendering...

PR #284, opened 2018-10-09
feat(popup-menu): make scaling configurable with default min and max

barmac:
Popup menu scaling is now globally configurable.

By default it does not scale below 1.0 and above 1.5.

Closes [#861](https://github.com/bpmn-io/bpmn-js/issues/861)

### DIFF
--- a/lib/features/popup-menu/PopupMenu.js
+++ b/lib/features/popup-menu/PopupMenu.js
@@ -16,16 +16,40 @@ import {
 var DATA_REF = 'data-id';
 
 
+function setTransform(element, transform) {
+  element.style['transform-origin'] = 'top left';
+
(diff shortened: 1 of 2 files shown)


lib/features/popup-menu/PopupMenu.js
```diff
@@ -282,6 +306,48 @@ PopupMenu.prototype._attachContainer = function(container, parent, cursor) {
...
+      scale = minScale;
+    }
+
+    if (isDefined(maxScale) && zoom > maxScale) {
+      scale = maxScale;
+    }
+
+  }
+
+  transform = 'scale(' + scale + ')';
```
nikku: I'd inline this statement directly rather than assigning to an additional variable:

```
setTransform(container, 'scale(...)');
```

nikku:
:+1:

PR #288, opened 2018-10-11
fix(auto-resize): resize collapsed element when it is expanded

### LINKED ISSUE #287: Resize collapsed element when it is expanded
__Describe the Bug__

The collapsed element does not resize after expansion even if the children have bigger size than the element. Resizing of a collapsed element is handled by `bpmn-js` (see: https://github.com/bpmn-io/bpmn-js/blob/master/test/spec/features/modeling/behavior/ToggleElementCollapseBehaviourSpec.js#L389) instead of `diagram-js`.

__Steps to Reproduce__

1. Create a collapsed element with a child of bigger size than the parent.
2. Expand the element.

__Expected Behavior__

The collapsed element should be resized.

__Environment__

 - Browser: Chrome 69
 - OS: MacOS Mojave
 - Library version: 2.6.1

barmac:
This PR fixes a problem which occurs when a collapsed element has children of bigger size than it is. Previously, the children would cover up the parent, making it unusable.

* AutoResize will now be activated whenever a collapsed element
  is expanded unless the auto resizing is explicitly deactivated
* (ReplaceShape) Hints will be now passed to create a replacing shape to determine
if auto resizing should be fired.
* AutoResize will be fired when a child element is resized
unless hints disallow it.

Closes #287 
Closes [#786](https://github.com/bpmn-io/bpmn-js/issues/786)

### DIFF
--- a/lib/features/auto-resize/AutoResize.js
+++ b/lib/features/auto-resize/AutoResize.js
@@ -94,6 +94,22 @@ export default function AutoResize(eventBus, elementRegistry, modeling, rules) {
 
     self._expand(shape.children || [], shape);
   });
+
+  this.postExecuted([ 'shape.resize' ], function(event) {
+    var context = event.context,
+        hints = context.hints,
+        shape = context.shape,
+        parent = shape.parent;
+
+    if (hints && (hints.root === false || hints.autoResize === false)) {
+      return;
+    }
+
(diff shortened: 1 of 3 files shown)


barmac:
Added commits resolving related problems with AutoResize

PR #290, opened 2018-10-24
WIP feat(selection-visuals): add marker to related elements

pinussilvestrus:
- adds the selection outline to related elements when
  - highlight label(s) when selecting/hovering sequence flow
  - highlight sequence flow when selecting/hovering label
- adjust SelectionVisualsSpec, because old version did not test expected behavior, since the `.djs-outline` class is set all the time, regardless the shape is selected before or not

Related to https://github.com/camunda/camunda-modeler/issues/369

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -8,7 +8,9 @@
 }
 
 .djs-element.hover .djs-outline,
-.djs-element.selected .djs-outline {
+.djs-element.selected .djs-outline,
+.djs-element.related-selected .djs-outline, 
+.djs-element.related-hover .djs-outline {
   visibility: visible;
   shape-rendering: crispEdges;
   stroke-dasharray: 3,3;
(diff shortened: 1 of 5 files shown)


nikku (dismissed):
From the user perspective, which alternatives did you consider for implementing this feature?

The following shows a quick recording of the look and feel in bpmn-js:

foo

Two things immediately get apparent:

* It is not possible for a user to distinguish the actual selected element from the connected element. This is a major UX flaw and needs to be addressed.
* Highlighting the label that belongs to an element only works for connections. Why is that? Any element may reference external labels (cf. start event in the screencapture shown above).

From the implementation perspective I'd ask you not to patch the stuff into the `SelectionVisuals` but to provide this as a new feature (call it `highlight-related` or something). This way users may decide to use it (or not) and the code base remains maintainable.

I already addressed:

* fix commit message style (using feature name `selection` as context)
* correct import to relative path (https://github.com/bpmn-io/diagram-js/pull/290/commits/6a5e6d049792e94252964b0c7e6ad568a043c962); importing from an absolute `lib` path within the `lib` directory works in tests only but fails in the final bundle

pinussilvestrus:
@nikku Thanks for feedback! Let's discuss the points you've mentioned

> It is not possible for a user to distinguish the actual selected element from the connected element. This is a major UX flaw and needs to be addressed

I've just make the opacity of the related element's outline a little less to make clear, which element is actually selected. I discussed it with others and I think it's fits for the user's perspective. What do you think? Sure, we can discuss coloring the related element boxes with another color or giving them another shape, but this would break the current experience a little bit.

image

> Highlighting the label that belongs to an element only works for connections. Why is that? Any element may reference external labels (cf. start event in the screencapture shown above).

Done!

> I'd ask you not to patch the stuff into the SelectionVisuals but to provide this as a new feature (call it highlight-related or something).

I've just created a new feature inside the `SelectionModule` right now, since creating a completely new module `highlight-related` would follow to that we have to adjust the module-loading in the `BpmnViewer` for activating. For such a small feature I would not do that, since it belongs to `selection` in some way.

nikku:
Solution sketch we'd like to try: 

image

PR #292, opened 2018-10-25
chore(project): detect unresolved imports

nikku:
This ensures we don't accidently require files in the lib directory
via the global `lib` import. That stuff works during tests but does
not work in the final bundle.

Considerations: This plugin adds some runtime overhead to our linting
process.

### DIFF
--- a/lib/.eslintrc
+++ b/lib/.eslintrc
@@ -1,0 +1,8 @@
+{
+  "plugins": [
+    "import"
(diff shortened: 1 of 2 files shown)


nikku:
Not to reviewer: The test failure on CI is a prove that this plugin works.

barmac:
Resolved merge conflict.

PR #294, opened 2018-10-26
376 add keyboard shortcuts to move elements

barmac:
This PR moves keyboard navigation with arrows to a separate feature as well as adds a `MoveSelection` feature which allows to move selected elements with keyboard. 

Related https://github.com/bpmn-io/bpmn-js/issues/376

This PR does not close the related issue as the feature still needs to be included in `bpmn-js`.


nikku:
Please rebase on top of the master branch.

PR #295, opened 2018-10-26
Editor action refactoring

nikku:
Refactors the editor actions to be independent from actual editor (or viewer) features, as discussed with @barmac.

Core features:

* only bind to actions in `EditorActions` which are implemented via features
* only register keyboard shortcuts for actions that are implemented via features
* rename `move-selection` to `keyboard-move-selection` for clarity
* adjust keyboard move selection config for clarity
* provide `EditorAction` Api for move selection feature
* expose key event via `keyEvent` property

Additional Benefit:

* EditorActions and Keyboard may now safely be used without pulling in the whole modeling code base.


barmac:
Reworded the refactor commits (`chore` => `fix`), squashed a step fix into the refactor commit so that it works as intended.

PR #296, opened 2018-10-29
fix(core/EventBus): do not iterate over mutating array of listeners

### LINKED ISSUE #293: EventBus _invokeListeners mutates list of listeners during for loop
__Describe the Bug__

EventBus _invokeListeners mutates list of listeners during for loop causing listeners not get executed.

__Steps to Reproduce__

We have a custom listener on commandStack.changed with a priority of 0.
One of the listeners for the event has a wrappedCallback that calls removeListener on itself.
The wrappedCallback listener is added when the replacement popup menu is opened here is the stack:
image

This removes a item from the list of listeners making the last listener not getting called at all.

eventbusbug

__Expected Behavior__

The last listener should still be executed.

__Workaround__

Add priority on custom event so it is not last.

__Environment__

 - diagram-js 2.5.1

nickeeex:
This pr fixes the problem of a run once listener removing itself from the list of listeners that can cause some listeners not to be called at all.

Closes #293

### DIFF
--- a/lib/core/EventBus.js
+++ b/lib/core/EventBus.js
@@ -332,9 +332,14 @@ EventBus.prototype._invokeListeners = function(event, args, listeners) {
 
   var idx,
       listener,
-      returnValue;
+      returnValue,
+      listenersCopy;
 
-  for (idx = 0; (listener = listeners[idx]); idx++) {
+  // make a shallow copy of the listeners
+  // due to possible mutation of the listeners array
+  listenersCopy = listeners.slice(0);
+
+  for (idx = 0; (listener = listenersCopy[idx]); idx++) {
 
     // handle stopped propagation
     if (event.cancelBubble) {
--- a/test/spec/core/EventBusSpec.js
+++ b/test/spec/core/EventBusSpec.js
@@ -523,6 +523,31 @@ describe('core/EventBus', function() {
       expect(listener1).not.to.have.been.called;
     });
 
+    it('should call all listeners atleast once', function() {
+      // given
+      var listener = sinon.spy();
+      var listener2 = sinon.spy();
+
+      eventBus.once('foo', listener);
+      eventBus.on('foo', listener2);
+
+      // when
+      eventBus.fire('foo');
+
+      // then
+      expect(listener).to.have.been.calledOnce;
+      expect(listener2).to.have.been.calledOnce;
+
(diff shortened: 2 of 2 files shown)


lib/core/EventBus.js
```diff
@@ -332,9 +332,14 @@ EventBus.prototype._invokeListeners = function(event, args, listeners) {
 
   var idx,
       listener,
-      returnValue;
+      returnValue,
+      listenersCopy;
+
+  // make a shallow copy of the listeners
+  // due to possible mutation of the listeners array
+  listenersCopy = listeners.slice(0);
```
nikku: Creating a shallow copy may cause quite a performance penalty. A better solution could be to keep the listener index an instance property and adjust it once removing `only` listeners.

nickeeex: Would something like this be better if slice is possibly too slow
``` 

  for (idx = 0; (listener = listeners[idx]);) {

    // handle stopped propagation
    if (event.cancelBubble) {
      break;
    }
    var oldLength = listeners.length;
    returnValue = this._invokeListener(event, args, listener);
    if (oldLength === listeners.length) {
      idx++;
    }
  }

**Comment by nikku** [comment]
Back from holidays, sorry for the delay.

**Comment by nikku** [comment]
Looking a bit deeper into the issue it seems like this is more complicated then expected.

Basic example:

```
eventBus.on('a', beforeFn);
eventBus.once('a', onceFn);
eventBus.once('a', otherOnceFn);
eventBus.on('a', afterFn);
```

This should work and call `beforeFn`, `onceFn`, `otherOnceFn`, `afterFn`.

But what about the following:

```
eventBus.once('a', function() {
  eventBus.once('a', LOW_PRIORITY, onceFn);
});
eventBus.on('a', afterFn);
```

This should technically execute `afterFn`, `onceFn`, too, right?

Or what about this one?

```
eventBus.once('a', function() {
  eventBus.once('a', HIGHER_PRIORITY, onceFn);
});
eventBus.on('a', afterFn);
```

This should _not_ execute `onceFn`, as it is added with higher priority than the current one right?

nikku:
The solution I went for is implementing the listener execution in a position independent way (not iterating index based but through a double linked list): https://github.com/bpmn-io/diagram-js/pull/308.

PR #306, opened 2018-12-17
Allow pasting of labels

### LINKED ISSUE #206: Label is not being pasted
`CopyPaste` currently copies labels but does not paste them due to [this line](https://github.com/bpmn-io/diagram-js/blob/master/lib/features/modeling/cmd/PasteHandler.js#L146).

__Expected__

* [ ] Copy paste pastes the label with the shape
* [ ] There exist test cases verifying the proper label copy and paste interaction

Broken test case attached via https://github.com/bpmn-io/diagram-js/commit/4d5c744d5aa2afc1613037fb089b66062bb466d3.

gustavjf:

Closes #206

Includes passing reproduction test case from 4d5c744.

I'm still familiarizing myself with the library so perhaps I've inadvertently made a mistake. The solution is borrowed from https://github.com/bpmn-io/diagram-js/issues/206#issuecomment-281365896 but considers multi-label support (i.e. `label` -> `labels` as an array).

Happy to make any necessary changes.

### DIFF
--- a/lib/features/modeling/cmd/PasteHandler.js
+++ b/lib/features/modeling/cmd/PasteHandler.js
@@ -179,13 +179,12 @@ PasteHandler.prototype.postExecute = function(context) {
 
   forEach(labels, function(labelDescriptor) {
     var labelTarget = self._getCreatedElement(labelDescriptor.labelTarget, tree),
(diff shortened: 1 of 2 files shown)


gustavjf:
Closed as the solution is more of a workaround.

PR #307, opened 2019-01-08
test(project): use built-in progress reporter

nikku:
* up to 30% faster test execution over spec reporter


barmac:
Updated `package-lock.json`.

PR #308, opened 2019-01-08
chore(core/EventBus): implement as double linked list

### LINKED ISSUE #293: EventBus _invokeListeners mutates list of listeners during for loop

nikku:
Closes #293

### DIFF
--- a/lib/core/EventBus.js
+++ b/lib/core/EventBus.js
@@ -150,7 +150,6 @@ EventBus.prototype.on = function(events, priority, callback, that) {
 
   events.forEach(function(e) {
     self._addListener(e, {
-      previous: null,
       priority: priority,
       callback: actualCallback,
       next: null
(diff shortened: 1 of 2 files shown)


barmac:
Removed never read `previous` property so it's a single linked list now.
Restructured `EventBus` tests so that the suite looks like this:
```javascript
describe('core/EventBus', function() {
  describe('feature', function() {
    it(...)
  });
});
```

PR #320, opened 2019-04-18
fix(modeling/reconnect): correctly layout endpoints

barmac:
This PR corrects layouting behavior on reconnection.

### DIFF
--- a/lib/features/modeling/cmd/ReconnectConnectionHandler.js
+++ b/lib/features/modeling/cmd/ReconnectConnectionHandler.js
@@ -15,9 +15,7 @@ ReconnectConnectionHandler.prototype.execute = function(context) {
   var newSource = context.newSource,
       newTarget = context.newTarget,
       connection = context.connection,
-      dockingOrPoints = context.dockingOrPoints,
-      oldWaypoints = connection.waypoints,
-      newWaypoints;
+      dockingOrPoints = context.dockingOrPoints;
 
   if (!newSource && !newTarget) {
     throw new Error('newSource or newTarget are required');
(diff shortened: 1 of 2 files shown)


lib/features/modeling/cmd/ReconnectConnectionHandler.js
```diff
@@ -29,38 +27,36 @@ ReconnectConnectionHandler.prototype.execute = function(context) {
...
-      connectionStart = getDocking(connection.waypoints[0]),
-      connectionEnd = getDocking(connection.waypoints[connection.waypoints.length - 1]);
+      dockingOrPoints = context.dockingOrPoints,
+      newSource = context.newSource,
+      movedEnd = newSource ? 'connectionStart' : 'connectionEnd',
+      newWaypoint,
+      hints = {};
+
+  if (isArray(dockingOrPoints)) {
+    connection.waypoints = dockingOrPoints;
```
nikku: We cannot leak this side effect in post execute. We must handle it (and properly revert it) in execute.

test/spec/features/modeling/ReconnectConnectionSpec.js
```diff
@@ -106,8 +108,12 @@ describe('features/modeling - reconnect connection', function() {
 
       it('should execute', inject(function(modeling) {
 
+        // given
         var newWaypoints = [ { x: 110, y: 110 }, { x: 300, y: 300 } ];
 
+        // stub layouter to not interfere with results;
+        sinon.stub(modeling, 'layoutConnection').returns();
```
nikku: I'd love to do less spying here so we keep our test as integration-testish as possible. If you do not want a certain behavior in your tests you can simply stub out the whole component (cf. https://github.com/bpmn-io/diagram-js/pull/320/commits/11baa5ce9515c4947eabf71b65dc63c9bb4d2f36) how I did it.

test/spec/features/modeling/ReconnectConnectionSpec.js
```diff
@@ -212,6 +241,23 @@ describe('features/modeling - reconnect connection', function() {
         expect(connection.waypoints).to.eql(oldWaypoints);
       }));
 
+
+      it('should layout connection', inject(function(modeling) {
```
nikku: I like that you added these explicit layout tests :tada:.

What about the `redo` case though? It would have made clear that the reconnect handler does not properly revert things on undo :wink:.

PR #327, opened 2019-05-02
fix(example): add source and target for example connection

pinussilvestrus:
It seems that the example leads to confusion in some place since when moving one of the shapes (visually) connected by the example connection, the connection didn't get updated (cf. [forum thread](https://forum.bpmn.io/t/diagram-js-with-connectors/3360)):

May-02-2019 09-33-12

This is _correct_ by design since the connection didn't get a `target` nor a `source` on creation. This pull request adds it to avoid confusion.

May-02-2019 09-36-11

### DIFF
--- a/example/app/app.js
+++ b/example/app/app.js
@@ -91,7 +91,9 @@ var connection1 = elementFactory.createConnection({
   waypoints: [
     { x: 250, y: 180 },
     { x: 290, y: 220 }
-  ]
+  ],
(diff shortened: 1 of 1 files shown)


pinussilvestrus:
If you think this is expected behavior, we could just leave it is at it is. In my opinion, it just confuses people, since newly created connection via the example ContextPad automatically align correctly, but the manually created doesn't.

PR #338, opened 2019-05-14
Consistently layout connections previews

barmac:
This unifies connection preview features already implemented in several components (Bendpoints, Create, Connect/GlobalConnect) within ConnectionPreview. 

Related https://github.com/bpmn-io/bpmn-js/issues/744

### DIFF
--- a/lib/features/bendpoints/BendpointMovePreview.js
+++ b/lib/features/bendpoints/BendpointMovePreview.js
@@ -84,7 +84,7 @@ export default function BendpointMove(injector, eventBus, canvas, graphicsFactor
   eventBus.on([
     'bendpoint.move.out',
     'bendpoint.move.cleanup'
-  ], function(e) {
+  ], HIGH_PRIORITY, function(e) {
 
     // remove connect marker
     // if it was added
--- a/lib/features/bendpoints/BendpointMovePreview.js
+++ b/lib/features/bendpoints/BendpointMovePreview.js
@@ -1,0 +1,159 @@
+import {
+  classes as svgClasses,
+  remove as svgRemove
+} from 'tiny-svg';
+
+import {
+  addBendpoint
+} from './BendpointUtil';
+
+import { translate } from '../../util/SvgTransformUtil';
+
+
+var MARKER_OK = 'connect-ok',
+    MARKER_NOT_OK = 'connect-not-ok',
+    MARKER_CONNECT_HOVER = 'connect-hover',
+    MARKER_CONNECT_UPDATING = 'djs-updating';
+
+var COMMAND_RECONNECT_START = 'connection.reconnectStart',
+    COMMAND_RECONNECT_END = 'connection.reconnectEnd';
+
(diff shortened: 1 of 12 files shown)


lib/features/bendpoints/BendpointMovePreview.js
```diff
@@ -23,20 +23,12 @@ var HIGH_PRIORITY = 1100;
 /**
  * A component that implements moving of bendpoints
  */
-export default function BendpointMove(injector, eventBus, canvas, graphicsFactory) {
-
-  // optional components
-  var connectionDocking = injector.get('connectionDocking', false),
-      layouter = injector.get('layouter', false);
+export default function BendpointMove(injector, eventBus, canvas) {
```
philippfromme: Rename to BendpointMovePreview.

lib/features/bendpoints/BendpointSnapping.js
```diff
@@ -7,7 +7,7 @@ import {
 var abs= Math.abs,
     round = Math.round;
 
-var TOLERANCE = 10;
+var TOLERANCE = 5;
```
philippfromme: This is problematic. We shouldn't have to care if grid snapping exists or not.

barmac: I think we should make bendpoint snapping totally optional and then remove it from bpmn-js. What's your opinion on that?

lib/features/bendpoints/BendpointMovePreview.js
barmac:
```suggestion
export default function BendpointMovePreview(injector, eventBus, canvas) {
```

lib/features/bendpoints/BendpointMovePreview.js
```diff
@@ -164,13 +145,14 @@ export default function BendpointMove(injector, eventBus, canvas, graphicsFactor
...
     }
 
-    redrawConnection(e);
+    if (connectionPreview) {
+      connectionPreview.cleanUp(context);
+    }
   });
 }
 
 BendpointMove.$inject = [
```
barmac:
```suggestion
BendpointMovePreview.$inject = [
```

philippfromme:
@barmac Is previewing while moving shapes not supposed to work?

barmac:
Ready again.

lib/features/bendpoints/BendpointSnapping.js
philippfromme: I don't think we should deactivate bendpoint snapping. Instead we should use `setSnapped` to mark the event as snapped if bendpoint snapping decides to snap it so it won't be snapped by grid snapping again.

If we decativate bendpoint snapping you can't snap to existing bendpoints with higher priority if they are not on the grid:

barmac: Cool. I'm reverting the commit about separation of features.

barmac:
@philippfromme feel free to merge this

PR #343, opened 2019-05-22
chore(create): merge multiple if statements

pinussilvestrus:
* follow-up of the 22-05-19 hour of code
* cf. https://github.com/bpmn-io/hour-of-code/blob/master/snippets/Create_canCreate.md


pinussilvestrus:
Also wondered about it, but it seems like we're using `chore` most of the times when we were refactoring stuff.

PR #345, opened 2019-05-23
fix(geometry): reverse truthy return values of `pointsAligned()`

gustavjf:
This was returning 'v' for horizontally-aligned lines and 'h' for vertically-aligned lines.

Added missing tests for `pointsAligned()`.

Also takes care of the width/height of the yellow parallel dragger SVG element and the direction of the dragger cursor when hovering over a line's parallel dragger area.

### DIFF
--- a/lib/features/bendpoints/BendpointUtil.js
+++ b/lib/features/bendpoints/BendpointUtil.js
@@ -112,7 +112,7 @@ function createParallelDragger(parentGfx, position, alignment) {
 
   svgAppend(draggerGfx, hit);
 
-  rotate(draggerGfx, alignment === 'h' ? 90 : 0, 0, 0);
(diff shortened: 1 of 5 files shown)


barmac:
It seems that `bpmn-js#LayoutConnectionBehavior#snapMiddleSegments` depends on the bug that this change fixes.

nikku:
Good catch.

nikku:
If we fix this we must double check and adapt our [dmn-js](https://github.com/bpmn-io/dmn-js) and [cmmn-js](https://github.com/bpmn-io/cmmn-js) editors accordingly, too.

gustavjf:
@nikku Had a look and `pointsAligned` is not used in those packages. Either way, I double-checked dmn-js and cmmn-js with linked diagram-js by running all tests and everything is green.

nikku:
Thanks :rocket:

PR #346, opened 2019-05-27
fix(bendpoints): correctly set initial context

barmac:


nikku:
After approving this I'm not 100% sure anymore why this is needed. Let's check that out tomorrow :speech_balloon:

PR #347, opened 2019-05-27
Allow setting and sharing of hints between create and commandStack.shape.create events

gustavjf:
This is useful for two cases:

1. Accessing and setting properties on the hints object in event handlers for create commands.

2. Sharing metadata between both event types.

### DIFF
--- a/lib/features/create/Create.js
+++ b/lib/features/create/Create.js
@@ -271,7 +271,8 @@ export default function Create(
         shape: shape,
         context: {
           shape: shape,
-          source: source
+          source: source,
+          hints: {}
         }
       }
     });
--- a/lib/features/create/Create.js
+++ b/lib/features/create/Create.js
@@ -13,6 +13,8 @@ import {
   remove as svgRemove
 } from 'tiny-svg';
 
+import { assign } from 'min-dash';
+
 import {
   translate
 } from '../../util/SvgTransformUtil';
@@ -217,6 +219,7 @@ export default function Create(
   eventBus.on('create.end', function(event) {
     var context = event.context,
         source = context.source,
+        hints = context.hints,
         shape = context.shape,
         target = context.target,
         canExecute = context.canExecute,
@@ -239,9 +242,12 @@ export default function Create(
       });
     } else {
       // invoke create, if connect is not set
-      shape = modeling.createShape(shape, position, target, {
-        attach: attach
-      });
+      shape = modeling.createShape(shape, position, target, assign(
+        {},
+        hints, {
+          attach: attach
+        }
+      ));
     }
 
     // make sure we provide the actual attached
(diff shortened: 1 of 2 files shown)


lib/features/create/Create.js
```diff
@@ -239,9 +242,11 @@ export default function Create(
       });
     } else {
       // invoke create, if connect is not set
-      shape = modeling.createShape(shape, position, target, {
-        attach: attach
-      });
+      shape = modeling.createShape(shape, position, target, assign(
+        hints, {
```
nikku: Please create a new, merged object safely by passing an empty object as the first argument: `assign({}, hints, { ... })`.

nikku: Could we make hint passing consistent and add it in the `modeling#appendShape` case, too?

test/spec/features/create/CreateSpec.js
```diff
@@ -483,6 +483,34 @@ describe('features/create - Create', function() {
...
+
+    it('should pass hints to commandStack.shape.create event context', inject(
+      function(create, eventBus, dragging) {
+
+        // given
+        var spy = sinon.spy(function(event) {
+          expect(event.context.hints.foo).to.exist;
+        });
+
+        eventBus.once('create.start', 0, function(event) {
```
nikku: priority is an optional argument, you do not have to pass `0`. Or does it actually make a difference here?

test/spec/features/create/CreateSpec.js
```diff
@@ -483,6 +483,34 @@ describe('features/create - Create', function() {
...
+          event.context.hints.foo = 'foo';
+        });
+
+        eventBus.once('commandStack.shape.create.execute', 0, spy);
```
nikku: priority is an optional argument, you do not have to pass `0`. Or does it actually make a difference here?

test/spec/features/create/CreateSpec.js
gustavjf: No, you're right.

lib/features/create/Create.js
gustavjf: Yes we could, but context hint passing is different in [Modeling#appendShape](https://github.com/bpmn-io/diagram-js/blob/487ffd40b984d94e4d886db109a216cfe0442caa/lib/features/modeling/Modeling.js#L337-L339) compared to [Modeling#createShape](https://github.com/bpmn-io/diagram-js/blob/487ffd40b984d94e4d886db109a216cfe0442caa/lib/features/modeling/Modeling.js#L285). 

If we want to include all hints in `appendShape`'s context like we do in `createShape`, it will result in a breaking change. 

Is that something we want to handle in this PR?

nikku: How is it going to be a breaking change?

gustavjf: If we want hints to be consistent in `modeling#appendShape` as they are now in `modeling#createShape`, the `connection`, `connectionParent` and `attach` properties would no longer be available directly on `context` but on `context.hints`.

So any code that calls something like `context.connection` will break after that change.

PR #349, opened 2019-05-28
Remove orange color from preview

barmac:
> ...because black is new orange.

This PR removes orange color from previews.

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -136,24 +136,6 @@ svg.new-parent {
   visibility: hidden;
 }
 
-/**
- * drag styles
- */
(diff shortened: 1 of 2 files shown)


nikku:
Let us consider this change not in isolation but together with all the other connection preview related improvements we have in mind.

Could you record before/after captures so we have a solid base for comparison of the different mechanism?

barmac:
Sure, I will do this.

barmac:
Element creation:

new-element-old 
new-element


Connections:

create-old 
create

barmac:
It's not gonna be merged, at least not yet.

PR #350, opened 2019-05-28
Improve parallel segment moving

gustavjf:
Parallel movement was previously possible only when hovering and clicking on the connection's center, within a hard-coded threshold of 10px.

The move region is two-thirds of a connection segment's length measured from the center of that segment.

The move activation threshold is now half of that region.

- [x] Ensure visual cue for segment move is shown on hover within move region
- [x] Ensure segment movement activation is triggered within move region
- [x] Ensure bendpoint creation is possible on start/end of segment
- [x] Ensure position of yellow segment dragger marker is the same as cursor position on segment hover (same as with bendpoint marker)
- [x] Ensure position of yellow segment dragger marker is the same as cursor position on `connectionSegmentMove.start`

Related to camunda/camunda-modeler#1197

### DIFF
--- a/lib/features/bendpoints/BendpointUtil.js
+++ b/lib/features/bendpoints/BendpointUtil.js
@@ -79,7 +79,7 @@ export function addBendpoint(parentGfx, cls) {
   return groupGfx;
 }
 
-function createParallelDragger(parentGfx, position, alignment) {
+function createParallelDragger(parentGfx, segmentStart, segmentEnd, alignment) {
   var draggerGfx = svgCreate('g');
 
   svgAppend(parentGfx, draggerGfx);
@@ -87,7 +87,7 @@ function createParallelDragger(parentGfx, position, alignment) {
   var width = 14,
       height = 3,
       padding = 6,
-      hitWidth = width + padding,
+      hitWidth = calculateHitWidth(segmentStart, segmentEnd, alignment, padding),
       hitHeight = height + padding;
 
   var visual = svgCreate('rect');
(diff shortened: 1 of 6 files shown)


pinussilvestrus:
Just some thoughts after we talked about the last thing which is open (cf. _segment dragger marker is on cursor position_): We could simply update the position of the marker on `element.mousemove` inside the `Bendpoints` module. Just like

```js
var draggerGfx = domQuery('.djs-segment-dragger', bendpointsGfx),
  visualGfx = domQuery('.djs-visual', draggerGfx);

svgAttr(visualGfx, {
   x: // actual position x
   y: // actual position y
});
```

gustavjf:
Thanks for the info @pinussilvestrus. That certainly makes sense but it seems like `element.mousemove` is never called once the cursor hovers within the new 2/3 hit area.

In fact, when the cursor enters the hit area, only `element.out`, `element.marker.update` `element.hover` and `element.marker.update` are called once, in that order. Nothing more.

I'm checking now to see why that is.

gustavjf:
Ok, I've looked into this further and `element.mousemove` is also never invoked when moving the cursor over the segment dragger before implementing the 2/3 hit region.

Moving the position of the segment dragger along with the cursor will require changing what events are fired in that situation. I'll have to investigate that further to see how that could work.

pinussilvestrus:
* Fixed the missing task (move segment dragger along with cursor)
* Refactor a bit and change test structure

Jun-12-2019 13-50-23

One last thing which is broken right now: On `connectionSegmentMove.start` it sets the dragger position back to the center:

Jun-12-2019 14-04-21

I'll have a look at this! 🍸

pinussilvestrus:
Fixed it accordingly (cf. 4497f8d)

Jun-18-2019 08-46-28

This is ready for review now. @barmac @nikku

PR #351, opened 2019-06-03
Remove redundant bendpoints from preview

barmac:
This PR prevents such behavior:

redundant

### DIFF
--- a/lib/features/bendpoints/BendpointMove.js
+++ b/lib/features/bendpoints/BendpointMove.js
@@ -1,7 +1,6 @@
 import {
-  pointDistance,
-  pointsOnLine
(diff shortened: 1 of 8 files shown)


nikku:
How does the new, improved interaction look like? For me it looks like this, with bpmn-js `master` and your feature branch: 

foo

barmac:
@nikku This won't work for bpmn-js until https://github.com/bpmn-io/bpmn-js/compare/1048-Connection-layout-preview-breaks-on-bendpoint-move is merged. Please test on that branch or check out the tests.

barmac:
@philippfromme I moved the filtering to BendpointMove so that it's not forced for each connection preview.

PR #356, opened 2019-06-05
chore(styles): disable .new-parent overlay for frame elements

pinussilvestrus:
Disables the `.new-parent` overlay for frame elements.

Closes bpmn-io/bpmn-js#1054

Jun-05-2019 08-17-02

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -37,6 +37,10 @@
   fill: #F7F9FF !important;
 }
 
+.djs-frame.new-parent .djs-visual > :nth-child(1) {
+  fill: none !important;
+}
+
 svg.drop-not-ok {
   background: #f9dee5 /* light-red */ !important;
 }


assets/diagram-js.css
```diff
@@ -37,6 +37,10 @@
   fill: #F7F9FF !important;
 }
 
+.djs-frame.new-parent .djs-visual > :nth-child(1) {
```
nikku: Shouldn't we simply ignore hover visually by creating an appropriate bpmn-js rule: https://github.com/bpmn-io/diagram-js/blob/master/lib/features/create/Create.js#L183?

barmac: +1

pinussilvestrus: Good catch! But then, we will need to create a bpmn-rule for `canMove`, since the [`MovePreview`](https://github.com/bpmn-io/diagram-js/blob/master/lib/features/move/MovePreview.js#L162) sets the `.new-parent` marker in this case. Right now, we simply allow hovering labels over any elements (cf. [`BpmnRules`](https://github.com/bpmn-io/bpmn-js/blob/master/lib/features/rules/BpmnRules.js#L453)).

This way we would have to _disallow_ hover labels on Group elements which sets the `.drop-not-ok` marker in the `MovePreview`, since atm the MovePreview forces to set a marker in any case, and has no option to simply ignore the hover. It would look as follows:

Jun-05-2019 08-52-03

To summarize, we have the following options:

- Disallow hovering labels over groups via bpmn-js rule (cf. screencast)
- Modify the MovePreview in some way to ignore the hover visually
- Set css property for `.djs-frame` (cf. this PR)

I'd go for the first solution.

barmac: Same here. Groups should never be considered as a new parent to any element.

pinussilvestrus: I close this and open another one in bpmn-js.

pinussilvestrus:
Close this PR as a follow-up for [this discussion](https://github.com/bpmn-io/diagram-js/pull/356#discussion_r290607745)

assets/diagram-js.css
nikku: I'm wondering why we have [this](https://github.com/bpmn-io/bpmn-js/blob/8862865e2a6eafc884d61b1a3b02bdc2d5912c8b/lib/features/rules/BpmnRules.js#L454) in the first place. Usually we return _null_ in these cases to visually ignore the thing.

PR #362, opened 2019-06-12
Escape text shown in search pad

nikku:


nikku:
Backported to:

* `2.x`
* `3.x`

PR #363, opened 2019-06-13
feat(i18n): escape translations per default

barmac:
This makes sure that we escape translations per default to avoid XSS issues with i18n. Pass `safe=true` if the input is already safe.

### DIFF
--- a/lib/util/EscapeUtil.js
+++ b/lib/util/EscapeUtil.js
@@ -3,12 +3,17 @@ export {
 } from 'css.escape';
 
 var HTML_ESCAPE_MAP = {
-  '<': '&lt',
-  '>': '&gt'
+  '&': '&amp;',
+  '<': '&lt;',
+  '>': '&gt;',
+  '"': '&quot;',
+  '\'': '&#39;'
 };
 
 export function escapeHTML(str) {
-  return str.replace(/[<>]/g, function(match) {
+  str = '' + str;
+
+  return str && str.replace(/[&<>"']/g, function(match) {
     return HTML_ESCAPE_MAP[match];
   });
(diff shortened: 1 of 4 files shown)


lib/util/EscapeUtil.js
```diff
@@ -2,13 +2,20 @@ export {
...
+  '&': '&amp;',
+  '<': '&lt;',
+  '>': '&gt;',
+  '"': '&quot;',
+  '\'': '&#39;'
 };
 
 export function escapeHTML(str) {
-  return str.replace(/[<>]/g, function(match) {
+  str = '' + str;
```
nikku: Regexes are stateful per default. Can we inline the `UNSAFE_HTML` tag?

nikku: Why are we escaping the extra characters, too? What is the purpose?

barmac: > Regexes are stateful per default. Can we inline the `UNSAFE_HTML` tag?

Sure, I'll do this.

> Why are we escaping the extra characters, too? What is the purpose?

If unescaped, ampersand will swallow a few following characters. 

Regarding the quotation marks, they might be a problem if inserted as part of attribute. However, I haven't found any example of this in the libraries. I didn't look for it in all of them, though.

nikku: All right. Let's go for full escaping then and see what happens.

nikku:
Seems like I merged this one prematurely. 

Escaping translations breaks the library and tools building upon it: https://github.com/bpmn-io/bpmn-js/pull/1080. Let's discuss and revise this next week.

PR #368, opened 2019-06-19
Make hosts sticky for valid attachments

barmac:
The (potential) host's hit shape will now grow so that it gives a _stickiness_ effect. When merged, this will fix https://github.com/bpmn-io/bpmn-js/issues/1078.

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -176,6 +176,10 @@ marker.djs-dragger text {
   pointer-events: none;
 }
 
+.djs-element.attach-ok .djs-hit {
+  stroke-width: 60px !important;
+}
+
 /**
  * all pointer events for hit shape
  */
--- a/lib/features/attach-support/AttachSupport.js
+++ b/lib/features/attach-support/AttachSupport.js
@@ -17,10 +17,13 @@ import { getNewAttachShapeDelta } from '../../util/AttachUtil';
 
 import inherits from 'inherits';
 
+import CommandInterceptor from '../../command/CommandInterceptor';
+
+
 var LOW_PRIORITY = 251,
     HIGH_PRIORITY = 1401;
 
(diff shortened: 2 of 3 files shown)


assets/diagram-js.css
```diff
@@ -176,6 +176,20 @@ marker.djs-dragger text {
   pointer-events: none;
 }
 
+/**
+ * hit shape
+ */
+.djs-hit {
```
nikku: We cannot do this, as it essentially prevents users from interacting with the viewer without the CSS file.

assets/diagram-js.css
```diff
@@ -176,6 +176,20 @@ marker.djs-dragger text {
...
+  fill: none;
+  stroke: white;
+  stroke-opacity: 0.0;
+  stroke-width: 15px;
+}
+
+.djs-hit-extended .djs-hit {
```
nikku: We already apply a `attach-ok` marker / class to elements that are show the attach preview. Could we not make the extended hit shape a generic behavior? I.e. 

```
.djs-element.attach-ok .djs-hit {
  stroke-width: 80px;
}
```

nikku: I do not see a reason why not all editors should inherit the sticky behavior for attachers.

barmac: This makes sense.

assets/diagram-js.css
barmac: OK

barmac:
@nikku I rewrote this PR so that it actually fixes the bpmn-js problem within diagram-js. Please feel free to leave your comments.

nikku:
Updated with minor changes, cf. https://github.com/bpmn-io/diagram-js/pull/368/commits/aec89d739ca4ccc06d32d3fe7b2ca85a8cd69fed.

PR #372, opened 2019-06-25
Snap Bendpoint If Outside Of Target

barmac:
Related to https://github.com/bpmn-io/bpmn-js/issues/529

### DIFF
--- a/lib/features/snapping/ConnectionSnapping.js
+++ b/lib/features/snapping/ConnectionSnapping.js
@@ -1,0 +1,102 @@
+import {
+  getOrientation
+} from '../../layout/LayoutUtil';
+
+var HIGH_PRIORITY = 2000;
+
+var ORIENTATION_PADDING = -10,
+    SNAP_PADDING = 10;
+
+
+/**
+ * Snap connection event to target center if outside of the target
+ *
+ * @param {EventBus} eventBus
+ */
+export default function ConnectionSnapping(eventBus) {
+
+  var self = this;
+
+  eventBus.on([
+    'connect.hover',
+    'connect.move',
+    'connect.end'
+  ], HIGH_PRIORITY, function(event) {
+
+    var context = event.context,
+        target = context.target,
+        orientation;
+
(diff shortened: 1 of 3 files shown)


nikku:
How does it feel? Do you think it is safe to merge this one from the UX perspective? Or should we give it another round in future releases?

barmac:
So there are a few quirks, but overally it feels much better than the current state.

The quirks are as follows:

1) It is still possible to connect _outside_ of the shape due to rounded corners ignored by the `getOrientation` util.

Screenshot 2019-06-25 at 18 55 06

2) The preview bendpoint is drawn on the position of snapped event, which is in the center of the shape.

Screenshot 2019-06-25 at 18 58 44

nikku:
I gave it a try and, given the quirks you've mentioned, would suggest that we do not include this in this milestone. The current solution approach introduces a large amount of noise and is still not solid (with regards to rounded corners or non-rectangular shapes (i.e. gateways and tasks).

out qaUDP

barmac:
@nikku please have a look at https://github.com/bpmn-io/bpmn-js/pull/1109 

I rewrote this branch to more or less port the simple solution we already have in bpmn-js.

I left the previously mentioned intersection detection experiments on https://github.com/bpmn-io/diagram-js/tree/snap-to-intersection-experiments

nikku:
What about the quirks you've mentioned? Does this feel for you like it's ready to be merged and released?

barmac:
It's much better now, with the jumping nor weirdly positioned bendpoint position indicator.

See:

lgtm

barmac:
I added an issue for a quirk that it does not fix nor break: https://github.com/bpmn-io/bpmn-js/issues/1110

barmac:
So I think this can be merged along with the bpmn-js PR

nikku:
But is it universally usable, i.e. for data stores and gateways, too?

barmac:
Decided to close this until we know what the correct solution is.

PR #375, opened 2019-06-26
chore(CHANGELOG): update with breaking change

barmac:
We missed the reason why it's going to be a major version. This PR adds the breaking change to the changelog.


nikku:
To be honest we should probably do a dry run on upgrading DMN and CMMN to see if there exists more of such changes. Could you do that?

nikku:
Snapping seems to be another breaking change:

image

nikku:
In diagram-js we've introduced a large number of additional snap points secretly via https://github.com/bpmn-io/diagram-js/commit/8469f3cb270164406cbab580a93a77ba2164abcb.

We got to decide whether to undo that change (and give toolkit consumers more fine grained control) or market that as a feature.

barmac:
@nikku No test failures in dmn-js after `diagram-js@4.0.0-beta.6` installation.

nikku:
So it seems like the only breaking bit is the additional snap points that I propose to unbuild (cf. https://github.com/bpmn-io/bpmn-js/issues/1104). Thanks for investigating.

PR #391, opened 2019-08-02
Ensure firing out event

pinussilvestrus:
This ensures we definitely fire an `element.out` event between two `hover` for different elements. This way we make sure cursor and marker cleanup really got to run, despite the fact [some browsers swallow certain `mouseout` events](http://stackoverflow.com/questions/7448468/why-cant-i-reliably-capture-a-mouseout-event).

Related to camunda/camunda-modeler#1383

### DIFF
--- a/lib/features/dragging/HoverFix.js
+++ b/lib/features/dragging/HoverFix.js
@@ -6,20 +6,19 @@ import {
   toPoint
 } from '../../util/Event';
 
-function getGfx(target) {
-  var node = domClosest(target, 'svg, .djs-element', true);
-  return node;
-}
+var HIGH_PRIORITY = 1500;
 
 
 /**
- * Browsers may swallow the hover event if users are to
+ * Browsers may swallow certain events (hover, out ...) if users are to
  * fast with the mouse.
  *
  * @see http://stackoverflow.com/questions/7448468/why-cant-i-reliably-capture-a-mouseout-event
  *
  * The fix implemented in this component ensure that we
- * have a hover state after a successive drag.move event.
+ *
+ * 1) have a hover state after a successive drag.move event
+ * 2) have an out event when dragging leaves an element
  *
  * @param {EventBus} eventBus
  * @param {Dragging} dragging
@@ -29,15 +28,16 @@ export default function HoverFix(eventBus, dragging, elementRegistry) {
 
   var self = this;
 
-  // we wait for a specific sequence of events before
-  // emitting a fake drag.hover event.
-  //
-  // Event Sequence:
-  //
-  // drag.start
-  // drag.move
-  // drag.move >> ensure we are hovering
-  //
+  /**
+   * We wait for a specific sequence of events before
+   * emitting a fake drag.hover event.
+   *
+   * Event Sequence:
+   *
+   * drag.start
+   * drag.move
+   * drag.move >> ensure we are hovering
+   */
   eventBus.on('drag.start', function(event) {
 
     eventBus.once('drag.move', function() {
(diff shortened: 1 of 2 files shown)


pinussilvestrus:
This does not affect https://github.com/bpmn-io/bpmn-js/issues/1123 (despite the fact this bug may also be caused by a missing fired event)

pinussilvestrus:
> it emits a fake element.out event

+1 going for executing `drag.out` instead, should follow to the same result.

> it hooks into low-level DOM events to get the job done

This aims to ensure the additional out event is NOT thrown when it was already caught. But yeah true, I'd go for tracking the `element.out` instead of using the low-level DOM listener.

>  Is there a good reason why we don't combine the two into a single file?

I think there is not. I wanted to definitely separate the two fixes since they solve different things. Will go for having them in the same module but separate them there instead.

nikku:
Why do you expect an additional out event after hover was fired? And if it was thrown, would it hurt?

pinussilvestrus:
Not after the hover, but we have the following happy path in terms of events when we hover from one element to another:

`hover` Element A -> `out` Element A  -> `hover` Element B

The bug in camunda/camunda-modeler#1383 is caused by the fact the `out` event between the two hover's is not fired in some occasions. The fix aims to simply add this missing `out` if (and only if) it is not fired.

Why so (and does it hurt us if we have multiple out events)? Without the check, we run into side effects

Aug-02-2019 10-33-54

which look like bpmn-io/bpmn-js#1123, but it is happening _all the time_ after adding a connection between two elements (as shown in the screencast). We discussed it with @philippfromme yesterday to add the check to ensure, we will not run into this kind of side effects when triggering unnecessary out operations.

I'll update the PR to the [things mentioned above](https://github.com/bpmn-io/diagram-js/pull/391#issuecomment-517605785)

nikku:
Cool. Thanks for the clarification.

pinussilvestrus:
One additional finding

> it emits a fake element.out event

I tried to go for the [`drag.out`](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/dragging/Dragging.js#L327) way instead of emitting the `element.out` event. But at the point we are want to do the `out` operation for the `previousHovered` (Element A), the drag context is still set to Element B. This way it removes Element B from the drag context, which leads to side effects ([in fact failing test cases](https://github.com/bpmn-io/diagram-js/blob/develop/test/spec/features/attach-support/AttachSupportSpec.js#L523)). 

In fact, [waiting for the next `drag.hover`](https://github.com/bpmn-io/diagram-js/pull/391/files#diff-e7e763ff139315001b2d80af0f4e4c3dR39) does not follow to the exact event-order as described from me above. To really achieve this, we would have hook into _before_ a second  `drag.hover` is fired, but _after_ the mouse leaves Element A. This would follow to a similar solution like the current [`HoverFix`](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/dragging/HoverFix.js#L57). So we have two options

* NOT using `drag.out` but emitting `element.out` for Element A, leave the rest
* Changing the behavior to hook into _before_ a second `drag.hover` is fired

@nikku what do you think? I'd personally go for the first one.

pinussilvestrus:
Updated the PR to my preferred version.

nikku:
I'll need to look into the issue.

If you have specific requirements regarding the event order, does it make sense to bake these into the test cases? Right now the coverage you provided is very light. I'd assume we verify the actual order and occurrence of events, from hover to out:

```
expect(emittedEvents).to.eql([
  [ 'drag.hover', ... ],
  [ 'drag.out', ... ]
]);
```

pinussilvestrus:
Would totally go for additional test coverage, also for the existing `HoverFixSpec`. Originally planned to take the [existing test case](https://github.com/bpmn-io/diagram-js/blob/develop/test/spec/features/dragging/HoverFixSpec.js#L27) to the Hour of Code, since it does not test the overall functionality (despite the fact it's hard to simulate a MouseEvent, which would be needed in this case).

Really like the idea of testing the event order, since this would really ensure the expected behavior. Feel free to add additional thoughts [regarding the solutions](https://github.com/bpmn-io/diagram-js/pull/391#issuecomment-517627228) provided above if you had a deeper look.  We can also chat about it next week.

nikku:
I'm unsure about whether the things you've mentioned are issues I overlook or whether I came up with a solution that just works.

Could you checkout https://github.com/bpmn-io/diagram-js/commit/87f2b21ef0ca01f2fb2c7eb595c4ed0142670f3b and see if it suffers from the same limitations you mentioned?

Please add additional test coverage, regardless of the implementation we are going for. It should not depend on it, anyway. Hint: Do not do `dragging.setOptions({ manual: true })` and you can nicely script the drag interaction by simply emitting `element.hover` / `element.out` events.

pinussilvestrus:
The problem I described does not come out of your implementation because it just solved them 👍 Great approach! It ensures the things I mentioned (cf. `out` event for correct element) in a good manner I think ❤️ 

To complete this PR I will 
* [ ] add your changes to this 
* [ ] add additional test coverage

Thanks for your time!

pinussilvestrus:
Ready to review now!

After finishing, I was not able to reproduce bpmn-io/bpmn-js#1123 again. Could be just luck or we really tackled this problem here as well. We will have to double-check it then.

Also tried to improve the existing test in [`HoverSpec`](https://github.com/bpmn-io/diagram-js/blob/develop/test/spec/features/dragging/HoverFixSpec.js#L27) but it revealed as being very hard to mock proper mouse events for this case. Let's postpone this for a future Hour of Code.

PR #395, opened 2019-08-13
fix(modeling): prevent fractional coordinates after create

pinussilvestrus:
This ensures created elements will not have fractional coordinates after creating.

Closes bpmn-io/bpmn-js#1167

### DIFF
--- a/test/spec/features/modeling/CreateElementsSpec.js
+++ b/test/spec/features/modeling/CreateElementsSpec.js
@@ -212,6 +212,27 @@ describe('features/modeling - create elements', function() {
       expect(elements[0]).to.equal(shape);
     }));
 
+
+    it('should not have fractional coordinates', inject(function(elementFactory, modeling) {
+
+      // given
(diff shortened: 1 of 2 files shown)


test/spec/features/modeling/CreateElementsSpec.js
```diff
@@ -212,6 +215,28 @@ describe('features/modeling - create elements', function() {
...
+        x: 110,
+        y: 110,
+        width: 105,
+        height: 105
+      });
+
+      var elements = modeling.createElements(shape, position, parentShape);
+
+      // then
+      expect(pick(elements[0], ['x', 'y'])).to.eql({
```
nikku: You can use the `to.have.position` matcher here.

pinussilvestrus: 🤯

barmac: Cool stuff! :)

PR #398, opened 2019-08-19
Update eslint plugin bpmn io

barmac:
So this is how it looks like.


lib/draw/DefaultRenderer.js
```diff
@@ -28,6 +28,7 @@ var DEFAULT_RENDER_PRIORITY = 1;
  * @param {Styles} styles
  */
 export default function DefaultRenderer(eventBus, styles) {
+
   //
```
nikku: I like this line specially :lollipop:

PR #403, opened 2019-08-21
Fixup create

pinussilvestrus:
After merging the fixes from https://github.com/bpmn-io/diagram-js/pull/400 and https://github.com/bpmn-io/diagram-js/pull/399 pasting-without-moving was not possible inside the `rootElement` since the `target` got removed [via the next `create.move`](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/create/Create.js#L144). This got fired [without the updated dragContext](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/dragging/Dragging.js#L171) (it is using the old event context without the updated hover, therefore hover is still undefined in this case).

It is working fine inside container elements since the browser fires an additional `hover` event due to the `pointer-event` changes by setting and unsetting the `.djs-drag-active` class.

This pull request adds changes to fire another fake `hover` after the misleading `create.move`, to ensure the hover did not get lost. _Additionaly_ it ensures we remove markers despite the fact the target got removed in the `create.move` handler. 

This fixes the [broken test case](https://travis-ci.org/bpmn-io/diagram-js/builds/574734365#L828) on current master. Paste-without-moving should work perfectly fine now.

Kapture 2019-08-21 at 11 18 52

### DIFF
--- a/lib/features/create/Create.js
+++ b/lib/features/create/Create.js
@@ -167,11 +167,10 @@ export default function Create(
   });
 
   eventBus.on([ 'create.end', 'create.out', 'create.cleanup' ], function(event) {
-    var context = event.context,
-        target = context.target;
+    var hover = event.hover;
 
-    if (target) {
-      setMarker(target, null);
+    if (hover) {
+      setMarker(hover, null);
     }
   });
 
--- a/lib/features/dragging/HoverFix.js
+++ b/lib/features/dragging/HoverFix.js
@@ -39,12 +39,21 @@ export default function HoverFix(eventBus, dragging, elementRegistry) {
       return;
     }
 
-    var gfx = self._findTargetGfx(event);
+    var originalEvent = event.originalEvent;
+
+    var gfx = self._findTargetGfx(originalEvent);
 
     if (gfx) {
       var element = elementRegistry.get(gfx);
 
+      // 1) cancel current mousemove
+      event.stopPropagation();
+
+      // 2) emit fake hover for new target
       dragging.hover({ element: element, gfx: gfx });
+
+      // 3) re-trigger move event
+      dragging.move(originalEvent);
     }
   }
 
@@ -59,7 +68,7 @@ export default function HoverFix(eventBus, dragging, elementRegistry) {
    */
   eventBus.on('drag.start', function(event) {
 
-    eventBus.once('drag.move', function(event) {
+    eventBus.once('drag.move', HIGH_PRIORITY, function(event) {
 
       ensureHover(event);
 
@@ -124,15 +133,14 @@ export default function HoverFix(eventBus, dragging, elementRegistry) {
   });
 
   this._findTargetGfx = function(event) {
-    var originalEvent = event.originalEvent,
-        position,
+    var position,
         target;
 
-    if (!(originalEvent instanceof MouseEvent)) {
+    if (!(event instanceof MouseEvent)) {
       return;
     }
 
-    position = toPoint(originalEvent);
+    position = toPoint(event);
 
     // damn expensive operation, ouch!
     target = document.elementFromPoint(position.x, position.y);
--- a/test/spec/features/dragging/HoverFixSpec.js
+++ b/test/spec/features/dragging/HoverFixSpec.js
@@ -77,6 +77,41 @@ describe('features/dragging - HoverFix', function() {
       }
     ));
 
+
+    it('should trigger hover and then move', inject(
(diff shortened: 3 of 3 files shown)


lib/features/create/Create.js
```diff
@@ -168,7 +168,7 @@ export default function Create(
 
   eventBus.on([ 'create.end', 'create.out', 'create.cleanup' ], function(event) {
     var context = event.context,
-        target = context.target;
+        target = context.target || event.hover;
```
nikku: Can't we use `event.hover` here in the first place?

nikku:
> (it is using the old event context without the updated hover, therefore hover is still undefined in this case).

Can't we fix that behavior instead of working around it? Should be doable, right?

lib/features/create/Create.js
pinussilvestrus: Shouldn't break anything I guess...

pinussilvestrus:
You mean by [updating the event context after the `drag` has been executed](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/dragging/Dragging.js#L171)? Should be doable, but not sure whether it breaks something. Let's figure it out.

nikku:
Canceling the mouse move event and triggering it again would work, too.

nikku:
`move` --> not hovering --> `hover` + cancel current move --> `hovered` --> emit another move.

pinussilvestrus:
Nice idea! Sounds like a good approach.

pinussilvestrus:
@nikku worked this way, ready to review again.

test/spec/features/dragging/HoverFixSpec.js
```diff
@@ -11,6 +11,7 @@ import { createCanvasEvent as canvasEvent } from '../../../util/MockEvents';
 
 import dragModule from 'lib/features/dragging';
 
+var LOW_PRIORITY = 500;
```
nikku: I wonder why this is necessary. What is the behavior we'd like to support?

* default _move_ listeners are never called -> hook into `drag.move` with high priority
* default _move_ listeners are maybe called, depending on registration order -> keep implementation as is

My suggestion would be: 

* change the test case to verify that listeners with normal priority are called once only
* update the implementation (listen for `drag.move` with high priority)

lib/features/dragging/HoverFix.js
```diff
@@ -39,12 +39,21 @@ export default function HoverFix(eventBus, dragging, elementRegistry) {
...
       var element = elementRegistry.get(gfx);
 
+      // 1) cancel current mousemove
+      event.stopPropagation();
+
+      // 2) emit fake hover for new target
       dragging.hover({ element: element, gfx: gfx });
+
+      // 3) re-trigger move event
+      dragging.move(originalEvent);
```
nikku: This should also make listening for `create.hover` in `Create` to update the rules obsolete, right?

test/spec/features/dragging/HoverFixSpec.js
pinussilvestrus: Good catch, just updated this!

lib/features/dragging/HoverFix.js
pinussilvestrus: No, we still need it for container elements.

Yesterday we found out the browser fires an `out` event for the `root` and a `hover` event for the container element when after drag mode is initiated. Therefore the `new-parent marker` would be removed from the actual hover (the container element). This was the actual reason we added the `create.hover` listening.

test/spec/features/dragging/HoverFixSpec.js
```diff
@@ -77,6 +77,31 @@ describe('features/dragging - HoverFix', function() {
       }
     ));
 
+
+    it('should stop propagation', inject(
```
barmac: This test case does not necessarily test the order in which the events are fired. You can check out the commit that I planned to add on top before I noticed that review was redone ;)
https://github.com/bpmn-io/diagram-js/compare/fixup-create...fixup-create-test?expand=1

lib/features/dragging/HoverFix.js
nikku: Forgot that in the mean time, thanks for pointing it out.

test/spec/features/dragging/HoverFixSpec.js
pinussilvestrus: Thanks, just modified the test case with your bonus material 👍

PR #407, opened 2019-08-22
Exclude coverage from package

barmac:


barmac:
Build fails because `develop` fails.

nikku:
Apparently npm ignores what is part of `.gitignore`, even on `npm@6.10`.

PR #413, opened 2019-10-02
feat(create): allow attachment of shapes with labels on creation

oguzeroglu:
Related to #1208


lib/features/create/Create.js
```diff
@@ -55,6 +55,11 @@ export default function Create(
       return !element.parent;
     });
 
+    // ignore labels being created with their targets
+    elements = filter(elements, function(element) {
+      return !(isLabel(element) && elements.indexOf(element) > -1);
```
barmac: Is it possible that `elements.indexOf(element) > -1` is false in this case?

oguzeroglu: Ah yes, this should be element.labelTarget.

oguzeroglu:
@barmac ready!

PR #416, opened 2019-10-04
fix(core): check if gfx exists before prepending

oguzeroglu:
Related to https://github.com/bpmn-io/bpmn-js/pull/1212


oguzeroglu:
We decided the prependTo should crash if the gfx is undefined, the real issue here is that we should never have the gfx undefined, so the issue will be fixed within bpmn-js.

PR #421, opened 2019-10-15
Hide elements recursively

oguzeroglu:
Required by https://github.com/camunda/camunda-modeler/issues/1328

Kapture 2019-10-15 at 17 56 03

### DIFF
--- a/lib/features/modeling/cmd/ToggleShapeCollapseHandler.js
+++ b/lib/features/modeling/cmd/ToggleShapeCollapseHandler.js
@@ -1,3 +1,8 @@
+import {
+  assign,
+  forEach
+} from 'min-dash';
+
 /**
  * A handler that toggles the collapsed state of an element
  * and the visibility of all its children.
@@ -16,16 +21,16 @@ ToggleShapeCollapseHandler.prototype.execute = function(context) {
   var shape = context.shape,
       children = shape.children;
 
-  // remember previous visibility of children
-  context.oldChildrenVisibility = getElementsVisibility(children);
+  // recursively remember previous visibility of children
+  context.oldChildrenVisibility = getElementsVisibilityRecursive(children);
 
   // toggle state
   shape.collapsed = !shape.collapsed;
 
-  // hide/show children
-  setHidden(children, shape.collapsed);
+  // recursively hide/show children
+  var result = setHiddenRecursive(children, shape.collapsed);
 
-  return [shape].concat(children);
+  return [shape].concat(result);
 };
 
 
@@ -36,13 +41,13 @@ ToggleShapeCollapseHandler.prototype.revert = function(context) {
 
   var children = shape.children;
 
-  // set old visability of children
-  restoreVisibility(children, oldChildrenVisibility);
+  // recursively set old visability of children
+  var result = restoreVisibilityRecursive(children, oldChildrenVisibility);
 
   // retoggle state
   shape.collapsed = !shape.collapsed;
 
-  return [shape].concat(children);
+  return [shape].concat(result);
 };
 
 
@@ -55,26 +60,48 @@ ToggleShapeCollapseHandler.prototype.revert = function(context) {
  *
  * @return {Object}
  */
-function getElementsVisibility(elements) {
+function getElementsVisibilityRecursive(elements) {
 
   var result = {};
 
-  elements.forEach(function(e) {
-    result[e.id] = e.hidden;
+  forEach(elements, function(element) {
+    result[element.id] = element.hidden;
+
+    if (element.children) {
+      result = assign({}, result, getElementsVisibilityRecursive(element.children));
+    }
   });
 
   return result;
 }
 
 
-function setHidden(elements, newHidden) {
-  elements.forEach(function(element) {
+function setHiddenRecursive(elements, newHidden) {
+  var result = [];
+  forEach(elements, function(element) {
     element.hidden = newHidden;
+
+    result = result.concat(element);
+
+    if (element.children) {
+      result = result.concat(setHiddenRecursive(element.children, element.collapsed || newHidden));
+    }
   });
+
+  return result;
 }
 
-function restoreVisibility(elements, lastState) {
-  elements.forEach(function(e) {
-    e.hidden = lastState[e.id];
+function restoreVisibilityRecursive(elements, lastState) {
+  var result = [];
+  forEach(elements, function(element) {
+    element.hidden = lastState[element.id];
+
+    result = result.concat(element);
+
+    if (element.children) {
+      result = result.concat(restoreVisibilityRecursive(element.children, lastState));
+    }
   });
+
+  return result;
 }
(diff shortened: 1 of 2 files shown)


nikku:
Did you consider the following:

```
A
  -> B (collapsed)
     -> C
```

Collapse `A`.

oguzeroglu:
@nikku Yes this is considered, even exists in the gif. B keeps being collapsed and the visibility of C does not change (because it's under a collapsed parent).

I've also added a test coverage about it: 'expand do not show children under collapsed parents'

test/spec/features/modeling/ToggleShapeCollapseSpec.js
```diff
@@ -158,6 +208,22 @@ describe('features/modeling - toggle collapsed', function() {
...
+
+      it('expand do not show children under collapsed parents',
+        inject(function(modeling, commandStack) {
+
+          // given
+          modeling.toggleCollapse(shownChildShape);
+          modeling.toggleCollapse(shapeToCollapse);
+
+          // when
+          commandStack.undo();
```
nikku: This test case is fine, it does not cover the case [described here](https://github.com/bpmn-io/diagram-js/pull/421#issuecomment-543116802) though.

My previous comment in JS code:

```javascript
// given
modeling.toggleCollapse(shownChildShape);
modeling.toggleCollapse(shapeToCollapse);

// when
modeling.toggleCollapse(shapeToCollapse);

// then
// ?
```

lib/features/modeling/cmd/ToggleShapeCollapseHandler.js
```diff
@@ -16,16 +21,16 @@ ToggleShapeCollapseHandler.prototype.execute = function(context) {
...
+  // recursively remember previous visibility of children
+  context.oldChildrenVisibility = getElementsVisibilityRecursive(children);
 
   // toggle state
   shape.collapsed = !shape.collapsed;
 
-  // hide/show children
-  setHidden(children, shape.collapsed);
+  // recursively hide/show children
+  var result = setHiddenRecursive(children, shape.collapsed);
```
nikku: Did you consider a CSS only solution to hide the children?

oguzeroglu: No, I have not, but what setHidden method does is to set the hidden property of a shape, then GraphicsFactory set display:none to those having hidden = true. Could you explain further what do you mean by CSS only solution?

test/spec/features/modeling/ToggleShapeCollapseSpec.js
oguzeroglu: Yes, even though this test case has the same kind of behavior, I'll add another coverage for expanding instead of undoing. Besides that, in this test:
A -> shapeToCollapse
    B -> shownChildShape 
        C -> shownSubChildShape

We first collapse B, then collapse A, then undo which expands A and check if C is still invisible and B is still collapsed.

lib/features/modeling/cmd/ToggleShapeCollapseHandler.js
nikku: Given how the elements are structured in the DOM we could create a CSS class, lets say `djs-hidden` on the element group that hides all elements via `display: none` in the hierarchy. 

Not sure if it is a good approach, but maybe one we could consider.

oguzeroglu: Kapture 2019-10-18 at 10 30 10

Adding display:none to related djs-children class would be a solution (this does not hide direct editing related DOM elements though but this also can be fixed). However, IMO this creates an inconsistency regarding the Shape class, as the hidden property would still be false/undefined even though the shape itself is hidden. So here, setHiddenRecursive method ensures that the hidden property of all the children/subchildren are correctly set as well.

nikku: Thanks for investigating. 

> However, IMO this creates an inconsistency regarding the Shape class, as the hidden property would still be false/undefined even though the shape itself is hidden. 

Agreed. We should go for the proper, consistent solution.

lib/features/modeling/Modeling.js
```diff
@@ -505,6 +505,5 @@ Modeling.prototype.toggleCollapse = function(shape, hints) {
     shape: shape,
     hints: hints || {}
   };
-
```
nikku: Please do not include unrelated style changes.

lib/features/modeling/cmd/ToggleShapeCollapseHandler.js
```diff
@@ -55,26 +60,46 @@ ToggleShapeCollapseHandler.prototype.revert = function(context) {
...
+    }
   });
+
+  return result;
 }
 
-function restoreVisibility(elements, lastState) {
-  elements.forEach(function(e) {
-    e.hidden = lastState[e.id];
+function restoreVisibilityRecursive(elements, lastState) {
```
nikku: This does not properly handle expanding when a nested child element is collapsed, see the failing test case I've created: https://github.com/bpmn-io/diagram-js/pull/421/commits/2a9da7cffd92a0d1eb2c225b45d6c4ed7b432d6e

test/spec/features/modeling/ToggleShapeCollapseSpec.js
nikku: I've created a failing test case regarding this: https://github.com/bpmn-io/diagram-js/pull/421/commits/2a9da7cffd92a0d1eb2c225b45d6c4ed7b432d6e

oguzeroglu: Ah yes ok I see the mistake now. Included this case and made a fix. I initially considered this case then somehow unfixed this while refactoring. Thanks for the catch!

lib/features/modeling/cmd/ToggleShapeCollapseHandler.js
oguzeroglu: Ok I commited a fix about this. The restoreVisibilityRecursive actually works fine (undoing stuff), the problem was with the setHiddenRecursive method instead.

nikku:
Added another failing test case: https://github.com/bpmn-io/diagram-js/blob/2b82079cbb8543c21df05d5be593cdd3f2fe4acb/test/spec/features/modeling/ToggleShapeCollapseSpec.js#L254.

nikku:
That one, by the way models [this setup](https://github.com/bpmn-io/diagram-js/pull/421#issuecomment-543116802).

oguzeroglu:
@nikku I think there's a problem with the test you commited.  It's under the `describe('collapse',` test line 122. The beforeEach logic of this test is that the shapeToCollapse is initially not collapsed. So when we call toggleCollapse for shapeToCollapse in the beginning of the test, every children should be hidden but on the next line we expect shownSubChildShape.hidden to be false. Am I missing something here? I think the test is placed in the wrong block.

nikku:
You're right, fixed the test and everything is :green_heart: now.

PR #422, opened 2019-10-17
Restore selection correctly

barmac:
This PR filters out the non-existing elements when restoring selection on dragging cancelation.

Related to https://github.com/bpmn-io/bpmn-js/issues/1225

### DIFF
--- a/lib/features/create/Create.js
+++ b/lib/features/create/Create.js
@@ -239,14 +239,12 @@ export default function Create(
   }
 
   // cancel on <elements.changed> that is not result of <drag.end>
-  eventBus.on('elements.changed', cancel);
-
-  eventBus.on('create.end', HIGH_PRIORITY, function() {
-    eventBus.off('elements.changed', cancel);
-  });
-
-  eventBus.on('create.ended', function() {
+  eventBus.on('create.init', function() {
     eventBus.on('elements.changed', cancel);
+
+    eventBus.once([ 'create.cancel', 'create.end' ], HIGH_PRIORITY, function() {
+      eventBus.off('elements.changed', cancel);
+    });
   });
 
   // API //////////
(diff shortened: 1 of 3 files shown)


lib/features/create/Create.js
```diff
@@ -239,14 +239,12 @@ export default function Create(
...
 
   // cancel on <elements.changed> that is not result of <drag.end>
-  eventBus.on('elements.changed', cancel);
-
-  eventBus.on('create.end', HIGH_PRIORITY, function() {
-    eventBus.off('elements.changed', cancel);
-  });
-
-  eventBus.on('create.ended', function() {
+  eventBus.on('create.init', function() {
```
nikku: What is this change about? Is it related to https://github.com/bpmn-io/bpmn-js/issues/1225?

barmac: A slight refactoring to ensure we do not listen for `elements.changed` when no create action is ongoing. This is related to the PR as the fixed bug was introduced via https://github.com/bpmn-io/diagram-js/pull/420

nikku: Strictly speaking it is not [a refactoring](https://en.wikipedia.org/wiki/Code_refactoring) but a change/improvement of the existing behavior, right? I guess it's tested implicitly via the other test cases you've added?

barmac: This is already tested in https://github.com/bpmn-io/diagram-js/blob/61648c45cfdc8a5abcce48952836cfbfa7e9c750/test/spec/features/create/CreateSpec.js#L373

The effect of that piece of code is that instead of having a noop listener for almost the all time we listen for the events only when it's relevant.

nikku: Thanks for clarifying.

PR #430, opened 2019-11-06
Correctly unsubscribe auto-close handler

barmac:
This fixes a wrong listener being unsubscribed from events. Not sure though, if we should actually listen for context pad events in that place.

### DIFF
--- a/lib/features/popup-menu/PopupMenu.js
+++ b/lib/features/popup-menu/PopupMenu.js
@@ -16,6 +16,11 @@ import {
 
 var DATA_REF = 'data-id';
 
+var CLOSE_EVENTS = [
+  'contextPad.close',
+  'canvas.viewbox.changing',
+  'commandStack.changed'
+];
 
 /**
  * A popup menu that can be used to display a list of actions anywhere in the canvas.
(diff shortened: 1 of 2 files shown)


lib/features/popup-menu/PopupMenu.js
```diff
@@ -16,6 +16,11 @@ import {
 
 var DATA_REF = 'data-id';
 
+var CLOSE_EVENTS = [
+  'contextPad.close',
```
barmac: I left it but am not sure if we should really depend on the Context Pad here. Popup Menu can function correctly without that component.

barmac: Actually, in the case of `contextPad.close` I'd rather see that behaviour in the `bpmn-js` as this is the place where PopupMenu is coupled with Context Pad.

nikku: I agree. This should not belong to diagram-js. The popup menu has no idea that the context pad even exists.

nikku: On the other hand: We had that quirky behavior in here. So if there is no good place to put it in bpmn-js we should simply keep it here.

barmac: The good place I could think of would be the Replace Menu which is a popup that closes with the Context Pad being closed. Anyway, this is outside of scope of this PR.

PR #434, opened 2019-11-22
ZoomScroll - Ability to enable to use mouse wheel directly without ctrl key

mnp-mid:
I would like to provide some code for the ability to use the mouse wheel directly without the use of the ctrl key. This code does not change any current behavior but provides a flag to enable / disable the ability with default false. 

The variable names are all yours to change. It would be nice if you could take the code or provide similar ability. 

Found in:
https://forum.bpmn.io/t/zooming-or-vertical-scrolling-what-should-be-the-default-mouse-wheel-action/131

### DIFF
--- a/lib/navigation/zoomscroll/ZoomScroll.js
+++ b/lib/navigation/zoomscroll/ZoomScroll.js
@@ -27,6 +27,8 @@ var DELTA_THRESHOLD = 0.1;
 
 var DEFAULT_SCALE = 0.75;
 
+var DEFAULT_DIRECT_ZOOM_WITH_MOUSE_WHEEL = false;
+
 /**
  * An implementation of zooming and scrolling within the
  * {@link Canvas} via the mouse wheel.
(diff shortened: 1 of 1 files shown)


barmac:
The behaviour introduced by the PR was already removed via eb1cad884f078487572c58fe8fedbb557b3760a6. IMO if you want to restore it, it's better to prepare a plugin as we don't want to merge it again into the core.

nikku:
@barmac Is it easily possible to change the behavior via a plug-in? My guess is it is not, given the way ZoomScroll is currently implemented.

barmac:
> @barmac Is it easily possible to change the behavior via a plug-in? My guess is it is not, given the way ZoomScroll is currently implemented.

The only way I know would be to override `_handleWheel`.

nikku:
First of all, my appologies for keeping this PR hanging.

After some additional thought on this issue I came to the conclusion to not merge this PR. Two reasons:

* As @barmac indicated it is fairly straight forward to go back to the old (your proposed) behavior. 
* I cannot rule out at the moment whether this will cause conflicts in other interactions.

If you'd like this mode (zoom without CTRL key) to end up as a core feature, I suggest you to start by creating a feature request. From there on, we can investigate which places may be affected and how to best address this issue.

PR #441, opened 2020-01-29
fix(copy-paste) Copy hidden and collapsed properties when copy-pasting.

yT0n1:
__Addresses https://github.com/bpmn-io/bpmn-js/issues/1270__

Hidden and collapsed elements are now pasted correctly, however, three problems remain.
1. The preview is not generated correctly. 
2. Labels are pasted incorrectly and are not visible. However, this bug appears to happen because the parent reference is not set correctly, similar to https://github.com/bpmn-io/bpmn-js/issues/945. Because the label becomes the child of the root element, it is never unhidden, when its actual  collapsed parent is expanded. This issue also happens by just collapsing, moving and uncollapsing a subprocess, unrelated to copy-pasting. 
3. After paste the sequence flow is selectable, however, once the root element has been selected, it is no longer selectable and correctly inside the actual parent shape. 

pr

Please let me know which one of these bugs should be fixed in this PR and for which I should open additional issues. A pointer of where to fix 1 and 3 would also be welcome.

### DIFF
--- a/lib/features/copy-paste/CopyPaste.js
+++ b/lib/features/copy-paste/CopyPaste.js
@@ -139,6 +139,10 @@ export default function CopyPaste(
         descriptor[ property ] = element[ property ];
       }
     });
+
+    descriptor.hidden = element.hidden;
+    descriptor.collapsed = element.collapsed;
+
   });
 
   eventBus.on('copyPaste.pasteElements', function(context) {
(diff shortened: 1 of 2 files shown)


nikku:
Thanks for your PR. I'll address the additional issues in a follow up PR.

PR #452, opened 2020-03-27
Keep connections & attachments in place on host resize if possible

barmac:


lib/features/modeling/cmd/SpaceToolHandler.js
```diff
@@ -24,12 +32,20 @@ SpaceToolHandler.prototype.preExecute = function(context) {
...
+      // TODO(@barmac): replace with new Map() once we drop PhantomJS
+      oldBounds = new BoundsMap();
 
   // (1) move shapes
   this.moveShapes(movingShapes, delta);
 
-  // (2) resize shapes
+  // (2a) save old bounds of resized shapes
+  forEach(resizingShapes, function(shape) {
+    oldBounds.set(shape, getBounds(shape));
```
nikku: Proposal: Let's not use a map here but rather a plain old object that keys by element ID (`ID` => `oldBounds`).

barmac: Makes sense. I will do that.

barmac: OK done.

barmac:
@philippfromme @nikku Ready for review.

PR #454, opened 2020-04-03
chore(CHANGELOG): update to v6.5.0

barmac:


nikku:
I think there is more insight, right?

https://github.com/bpmn-io/diagram-js/compare/v6.4.1...develop

One thing on the list is autoplace, now added to diagram-js.

nikku:
> One thing on the list is autoplace, now added to diagram-js.

And this one is still work in progress.

We could decide to remove it again, as it will likely be a major change to have it in.

barmac:
Correct! I was only comparing to `master`...

barmac:
> > One thing on the list is autoplace, now added to diagram-js.
> 
> And this one is still work in progress.
> 
> We could decide to remove it again, as it will likely be a major change to have it in.

So then how about if I cherry-pick and leave https://github.com/bpmn-io/diagram-js/pull/443/commits on `develop`?

Issue #472, opened 2020-06-01
PopupMenu.open() method fires 'popupMenu.open' event before menu entries are attached to the DOM


suupca:
PopupMenu.open() method fires 'popupMenu.open' event before popup menu entries are attached to the DOM. Due to this a custom 'popupMenu.open' event listener is unable to find the custom popupMenu entry using document querySelector because my custom menu entry is not attached to the DOM yet. 

Proposing to move the 'this._emit('open');' line to the end of the PopupMenu.open() method after menu entries have been attached to the DOM.

__Steps to Reproduce__

1. Add custom popup menu options
2. add eventbus listener on 'popupMenu.open' event and try to read the custom menu entry using document query selector
3. The element is not found, whereas I expect the menu entry options to be available in the DOM.

__Environment__

Please complete the following information:

 - Browser: [e.g. IE 11, Chrome 69]
 - OS: [e.g. Windows 7]
 - Library version: [e.g. 2.0.0]

nikku:
What would you like to achieve?

suupca:
> 
> 
> What would you like to achieve?

proposed change: https://github.com/bpmn-io/diagram-js/pull/473

We would want to able to read the custom popup menu entry element using DOM selectors.

nikku:
Why?

PR #473, opened 2020-06-01
fix: moving this._emit('open') line to the end of open() method to al…

suupca:
PopupMenu.open() method fires 'popupMenu.open' event before popup menu entries are attached to the DOM. Due to this a custom 'popupMenu.open' event listener is unable to find the custom popupMenu entry using document querySelector because my custom menu entry is not attached to the DOM yet. 
Proposing to move the 'this._emit('open');' line to the end of the PopupMenu.open() method after menu entries have been attached to the DOM.

PR addresses this issue:
https://github.com/bpmn-io/diagram-js/issues/472


nikku:
What would you like to achieve?

PR #486, opened 2020-09-07
More detailed commandStack events

### LINKED ISSUE #479: Command stack change event should be more detailed.

PabloDons:
Closes #479 

I haven't rigorously tested it, but it seems stable for my usage. I have only used the execute method, but I went ahead and added undo, redo and clear to make it a complete feature.

### DIFF
--- a/lib/command/CommandStack.js
+++ b/lib/command/CommandStack.js
@@ -1,6 +1,7 @@
 import {
   uniqueBy,
-  isArray
+  isArray,
+  assign,
 } from 'min-dash';
 
 
@@ -107,10 +108,16 @@ export default function CommandStack(eventBus, injector) {
    * Current active commandStack execution
    *
    * @type {Object}
+   * @property {Array} actions
+   * @property {Array} dirty
+   * @property {boolean|undefined} atomic
+   * @property {string|undefined} method Is this an execution, undo, or redo?
+   * @property {Array} executed List of actions who finished executing
    */
   this._currentExecution = {
     actions: [],
-    dirty: []
+    dirty: [],
+    executed: [],
   };
 
 
@@ -141,6 +148,8 @@ CommandStack.prototype.execute = function(command, context) {
     throw new Error('command required');
   }
 
+  var execution = this._currentExecution;
+  execution.method || (execution.method = 'execute');
   var action = { command: command, context: context };
 
   this._pushAction(action);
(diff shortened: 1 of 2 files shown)


PabloDons:
Is this an issue with my code? I don't see any problems

lib/command/CommandStack.js
```diff
@@ -446,17 +461,21 @@ CommandStack.prototype._pushAction = function(action) {
...
 
   if (!actions.length) {
-    this._eventBus.fire('elements.changed', { elements: uniqueBy('id', dirty.reverse()) });
+    var context = { elements: uniqueBy('id', dirty.reverse()) };
+    this._eventBus.fire('elements.changed', assign({}, context));
 
     dirty.length = 0;
 
-    this._fire('changed');
+    this._fire('changed', assign({ method: method, actions: execution.executed }, { context }));
```
pinussilvestrus, line 476: I guess [the build is failing](https://travis-ci.org/github/bpmn-io/diagram-js/builds/724921967) because you use unsupported JS syntax. The shorthand object initializer is an es6 feature and therefore not available for older browsers.

pinussilvestrus:
Thanks a lot first for your contribution! We will discuss in the team how to handle your proposal. In the meantime, can you change your pull request structure to fit [our guidelines](https://github.com/bpmn-io/diagram-js/blob/develop/.github/CONTRIBUTING.md#creating-a-pull-request)? Especially test cases which test the new features and a formatted git commit message would be very appreciated.

test/spec/command/CommandStackSpec.js
```diff
@@ -876,4 +877,99 @@ describe('command/CommandStack', function() {
...
+        commandStack.redo();
+
+        // then
+        var changeEvent = eventData.changeEvent,
+            actions = eventData.actions;
+
+        expect(changeEvent).to.equal('redo');
+
+        // Need to figure out a way to indicate the executed actions
+        expect(actions).to.eql(['pre-command', 'complex-command', 'post-command']);
```
PabloDons, line 950: This one has an issue though, that I discovered when making tests (I skipped the relevant test because it fails). The redo action doesn't indicate the executed actions. I can't figure out a graceful way to do this, so I would appreciate a nudge in the right direction!

barmac:
I am closing this until we figure out what we actually want to have.

PabloDons:
I have amended the pull-request as I described in the related issue. Now it will only indicate the change method (`execute`, `undo`, or `redo`). I apologize for the delay.

I see however that the pull-request didn't update. Is this because it is closed?

nikku:
> I see however that the pull-request didn't update. Is this because it is closed?

Apparently you force pushed or updated the develop branch, that is why you cannot reopen the PR.

PR #487, opened 2020-09-07
feat(element-registry): add ElementRegistry#find method

### LINKED ISSUE #261: Add ElementRegistry#find method
We offer a number of collection-like utilities on the `ElementRegistry` already.

What is missing is the ability to find the first element that matches a certain condition.

Right now the workaround is:

```
var element = elementRegistry.filter(condition)[0];
```

...which will always iterate over the complete list of elements.

__Tasks__

* [ ] Provide `ElementRegistry#find` method

stdevi:
Closes #261

### DIFF
--- a/lib/core/ElementRegistry.js
+++ b/lib/core/ElementRegistry.js
@@ -138,6 +138,30 @@ ElementRegistry.prototype.filter = function(fn) {
 };
 
 /**
+ * Return the first element that satisfies the provided testing function.
+ *
+ * @param {Function} fn
+ *
+ * @return {djs.model.Base}
+ */
+ElementRegistry.prototype.find = function(fn) {
+
+  var map = this._elements,
+      keys = Object.keys(map);
+
+  for (var i = 0; i < keys.length; i++) {
+    var id = keys[i],
+        container = map[id],
+        element = container.element,
+        gfx = container.gfx;
+
+    if (fn(element, gfx)) {
+      return element;
+    }
+  }
+};
+
+/**
  * Return all rendered model elements.
  *
  * @return {Array<djs.model.Base>}
--- a/test/spec/core/ElementRegistrySpec.js
+++ b/test/spec/core/ElementRegistrySpec.js
@@ -206,6 +206,79 @@ describe('ElementRegistry', function() {
 
   });
 
+
+  describe('find', function() {
+
+    it('should find by id', inject(function(elementRegistry) {
+
+      // when
+      var element = elementRegistry.find(function(element) {
+
+        // assume we get element as params
+        expect(element).to.exist;
+
+        return element.id === '2';
+      });
+
+      // then
+      expect(element).to.exist;
+      expect(element.id).to.equal('2');
+    }));
+
+
+    it('should find by type', inject(function(elementRegistry) {
+
+      // when
+      var element = elementRegistry.find(function(element) {
+
+        // assume we get element as params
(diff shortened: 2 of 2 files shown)


lib/core/ElementRegistry.js
```diff
@@ -137,6 +137,29 @@ ElementRegistry.prototype.filter = function(fn) {
...
+ */
+ElementRegistry.prototype.find = function(fn) {
+
+  var map = this._elements;
+
+  for (var i = 0; i < Object.keys(map).length; i++) {
+    var id = Object.keys(map)[i];
+    var container = map[id],
+        element = container.element,
+        gfx = container.gfx;
```
pinussilvestrus, line 155: Better

```js
var id = Object.keys(map)[i],
       container = map[id],
       element = container.element,
       gfx = container.gfx;
```

lib/core/ElementRegistry.js
```diff
@@ -137,6 +137,29 @@ ElementRegistry.prototype.filter = function(fn) {
...
+ *
+ * @param {Function} fn
+ *
+ * @return {djs.model.Base}
+ */
+ElementRegistry.prototype.find = function(fn) {
+
+  var map = this._elements;
+
+  for (var i = 0; i < Object.keys(map).length; i++) {
```
pinussilvestrus, line 151: what about storing the keys in a separate variable, so you don't have to execute it twice?

```js
var keys = Object.keys(map)
```

test/spec/core/ElementRegistrySpec.js
```diff
@@ -206,6 +206,59 @@ describe('ElementRegistry', function() {
 
   });
 
+  describe('find', function() {
```
pinussilvestrus, line 209: Make sure you have _two new lines_ between any test-group (describe) and any test (it).

lib/core/ElementRegistry.js
stdevi: Agree, I will do that.

pinussilvestrus:
Thanks for your contribution! I especially like that you consider what the author of the issue (https://github.com/bpmn-io/diagram-js/issues/261) has said

> ...which will always iterate over the complete list of elements.

and prevent this by early return when one match was found. Can you also update your commit message to fit our guidelines?

```js
feat(ElementRegistry): add find method

Closes #261 
```

test/spec/core/ElementRegistrySpec.js
```diff
@@ -206,6 +206,62 @@ describe('ElementRegistry', function() {
 
   });
 
+
+  describe('find', function() {
```
pinussilvestrus, line 210: Should we maybe also add a test case, what happens, when no match was found?

lib/core/ElementRegistry.js
```diff
@@ -137,6 +137,30 @@ ElementRegistry.prototype.filter = function(fn) {
   return filtered;
 };
 
+/**
+ * Return the first elements that satisfies the provided testing function.
```
pinussilvestrus, line 141: "the first element" (no plural)

test/spec/core/ElementRegistrySpec.js
stdevi: Sure, I've added one additional test case.

pinussilvestrus:
thanks a lot! I merged it via cb45b735dd7ed24c28b9138ed4dd3f36baa84c11

PR #498, opened 2020-11-06
fix(keyboard): bind zoom in to `=` key

barmac:
This is required for the feature to work on international
keyboard layout.

Related to https://github.com/bpmn-io/bpmn-js/issues/1362

__Acceptance Criteria__

__Definition of Done__

### DIFF
--- a/lib/features/keyboard/KeyboardBindings.js
+++ b/lib/features/keyboard/KeyboardBindings.js
@@ -126,7 +126,9 @@ KeyboardBindings.prototype.registerBindings = function(keyboard, editorActions) 
 
     var event = context.keyEvent;
 
-    if (isKey([ '+', 'Add' ], event) && isCmd(event)) {
+    // quirk: it has to be triggered by `=` as well to work on international keyboard layout
+    // cf: https://github.com/bpmn-io/bpmn-js/issues/1362#issuecomment-722989754
+    if (isKey([ '+', 'Add', '=' ], event) && isCmd(event)) {
       editorActions.trigger('stepZoom', { value: 1 });
 
       return true;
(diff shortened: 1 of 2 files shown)


barmac:
@pinussilvestrus @nikku Can you please review this?

pinussilvestrus:
For my understanding (also cf. the discussion starting here: https://github.com/bpmn-io/bpmn-js/issues/1362#issuecomment-723000253):

This fix shouldn't cause that `CMD+=` triggers the zoom on German keyboards as well, right? Since this resolves to key `0`. 

That's at least what I'd expect and it looks like it's working. Don't have an international keyboard, so I can only verify it does not break the existing behavior on German keyboard.

nikku:
> This fix shouldn't cause that CMD+= triggers the zoom on German keyboards as well, right? Since this resolves to key 0.

Good point and makes me more confident in the solution suggested. I'll double check the behavior on Linux. Maybe we can give this a quick :+1: and merge.

barmac:
To sum up the concerns, this change could be a problem only in a keyboard layout where you can type `=` but it's not coupled with `+` on the same keyboard key. Only in that situation you could be surprised with the result of Cmd + [=]. I checked also AZERTY layout and it's not the case there.

pinussilvestrus:
> To sum up the concerns, this change could be a problem only in a keyboard layout where you can type `=` but it's not coupled with `+` on the same keyboard key. Only in that situation you could be surprised with the result of Cmd + [=]. I checked also AZERTY layout and it's not the case there.

So far I can't think about or even know a keyboard layout where this is the case. I think it's okay to merge this one (if it's working on all platforms) and wait for a response in case this breaks on any keyboard.

barmac:
I looked through Wikipedia and found some layouts I haven't heard of which could potentially run into this issue: JCUKEN (Latin), BÉPO and Blickensderfer. Still, for them it's just a second shortcut for them, but the first one for international (and Polish as well) layout users.

nikku:
Did not see any side-effects on Linux. 

Cannot reproduce the CTRL/CMD behavior (https://github.com/bpmn-io/bpmn-js/issues/1362#issuecomment-723003712), too. So that one seems to be a MacOS specific thing.

PR #499, opened 2020-11-24
Remove hover class of hovered element on drop

azeghers:
Related to https://github.com/bpmn-io/bpmn-js/issues/1366

__Acceptance Criteria__

__Definition of Done__

### DIFF
--- a/lib/features/dragging/HoverFix.js
+++ b/lib/features/dragging/HoverFix.js
@@ -24,7 +24,7 @@ var HIGH_PRIORITY = 1500;
  * @param {Dragging} dragging
  * @param {ElementRegistry} elementRegistry
  */
-export default function HoverFix(eventBus, dragging, elementRegistry) {
+export default function HoverFix(canvas, eventBus, dragging, elementRegistry) {
 
   var self = this;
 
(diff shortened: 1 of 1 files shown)


azeghers:
@nikku Let me know if this is the right direction and I'll write tests for it :smile:

nikku:
Please go ahead and write tests regardless of the actual implementation. Test should test the scenario at hand, using our APIs. They are often times more important than the actual implementation (where stuff is fixed and how), as they serve as the spec for the thing fixed.

nikku:
Will have a look later today.

lib/features/dragging/HoverFix.js
```diff
@@ -77,6 +77,16 @@ export default function HoverFix(eventBus, dragging, elementRegistry) {
 
   });
 
+  /**
+   * We make sure that dropping an element forces the element
+   * we're hovering over to lose the hover class.
+   */
+  eventBus.on('drag.end', function(event) {
+
+    canvas.removeMarker(event.hoverGfx, 'hover');
```
nikku, line 86: I thought the issue was:

Generally `element.hover { A }` -> `element.out {A}` -> `element.hover {B}` is fired
Firefox sometimes swallows the `element.out {A}` so we end up with a sequence of  `element.hover { A }` -> `element.hover {B}` .

My intuition is that we should detect that broken sequence and emit the missing `element.out {A}`. If that is possible, it would be my preferred option.

Your solution drops the hover class, good. But it still leaves other components hanging that may rely on the proper sequence of `hover -> out -> hover` events.

nikku:
Closed in favor of https://github.com/bpmn-io/diagram-js/pull/503.

PR #503, opened 2020-11-29
Generalize fix for swallowed hover events

azeghers:
Replaces the old drag hover fix with one that works in a general fashion only on element.* level.

Related to https://github.com/bpmn-io/bpmn-js/issues/1366

__Acceptance Criteria__

__Definition of Done__


lib/features/hover-fix/index.js
```diff
@@ -0,0 +1,14 @@
...
+
+export default {
+  __init__: [
+    'hoverFix'
+  ],
+  __depends__: [
+    DraggingModule
+  ],
+  hoverFix: [ 'type', HoverFix ],
+};
```
pinussilvestrus, line 14: Since we ship this as a standalone module, we have to integrate it into downstream libraries as `bpmn-js` separately, right?

test/spec/features/hover-fix/HoverFixSpec.js
```diff
@@ -8,11 +8,12 @@ import { assign } from 'min-dash';
 import { createCanvasEvent as canvasEvent } from '../../../util/MockEvents';
 
 import dragModule from 'lib/features/dragging';
+import hoverFixModule from 'lib/features/hover-fix';
 
 
 describe('features/dragging - HoverFix', function() {
```
pinussilvestrus, line 14:
```suggestion
describe('features/hover-fix - HoverFix', function() {
```

barmac:
> Closes bpmn-io/bpmn-js#1366

Generally, if we create a solution for an upstream issue, the fix does not solve the issue yet. It will be the update of the dependency which solves the problem and the fix can only be _Related to bpmn-io/bpmn-js#1366_. For example, have a look at https://github.com/camunda/camunda-modeler/pull/1997

barmac:
I've tested this on my machine and it works correctly when the new module is included in `bpmn-js`.

lib/features/hover-fix/index.js
nikku: @azeghers We decided to not break existing behavior and keep it included per default, right?

azeghers:
I have made the HoverFixModule a dependency of the DraggingModule, so that anything that imports it initializes the hover fix in the process. This should implicitly enable it everywhere, so long as the dragging module is used.

azeghers:
Additionally, the tests should now also ensure that the element-level hover fix doesn't need the dragging module bootstrapped to work, and that the dragging-level hover fix doesn't need the hover fix module bootstrapped to work.

test/spec/features/create/CreateSpec.js
```diff
@@ -27,7 +28,8 @@ var testModules = [
   attachSupportModule,
   modelingModule,
   moveModule,
-  dragModule
+  dragModule,
+  hoverFixModule
```
nikku, line 32: Not needed. (Included per default now.)

nikku (changes requested):
One last thing and we got it, I believe!

nikku (approved):

nikku:
:cool:

PR #508, opened 2020-12-10
feat(util/Text): support soft breaks

nikku:
This adds handling for discretionary hyphens / soft breaks in our renderer.

If a discretionary hyphen is rendered as a break point it becomes a `-`.

Related to https://github.com/bpmn-io/bpmn-js/issues/1383

---

capture KCFgUj_optimized

### DIFF
--- a/lib/util/Text.js
+++ b/lib/util/Text.js
@@ -275,7 +275,9 @@ Text.prototype.layoutText = function(text, options) {
 
   var lineHeight = getLineHeight(style);
 
-  var lines = text.split(/\r?\n/g),
+  // we split text by lines and normalize
+  // {soft break} + {line break} => { line break }
+  var lines = text.split(/\u00AD?\r?\n/),
       layouted = [];
 
   var maxWidth = box.width - padding.left - padding.right;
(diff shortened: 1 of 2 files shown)


lib/util/Text.js
```diff
@@ -133,21 +134,23 @@ function fit(lines, fitLine, originalLine, textBBox) {
  * @return {string} the shortened string
  */
 function semanticShorten(line, maxLength) {
-  var parts = line.split(/(\s|-)/g),
+
+  var parts = line.split(/(\s|-|­)/g),
```
azeghers, line 138: The regex doesn't look like it contains a special character at the end, it could inadvertently be removed by a future maintainer. I think this line deserves a comment that warns that a soft hyphen is present in the regex.
Or alternatively, you can match it using the unicode value `\u00AD`.

nikku: Agreed. Good catch.

nikku: Using explicit unicode char now. Much better: https://github.com/bpmn-io/diagram-js/pull/508/commits/3981c21ae4c79ca47526636376cd1c6e13f84774

azeghers:
LGTM
Commits should be squashed before merging, right?

azeghers (approved):

PR #512, opened 2020-12-16
Revert bendpoint preview quick fix

nikku:
This reverts https://github.com/bpmn-io/diagram-js/commit/66b77964ba692dc7da5454d7855a9043de88f89b. The fix looked simple enough but actually introduced some [unwanted side-effects](https://travis-ci.com/github/bpmn-io/bpmn-js/builds/208964265).

Specifically we do rely on `connectionPreview` to attach `getConnection` to the bendpoint move context in our tests, both in diagram-js as well as bpmn-js. 

While this is shaky in the first place (relying on an internal just like that) I'd rather revert the fix now and investigate an appropriate solution in an orderly manner.

I'll investigate if I can find a more appropriate solution.

### DIFF
--- a/lib/features/bendpoints/BendpointMovePreview.js
+++ b/lib/features/bendpoints/BendpointMovePreview.js
@@ -108,12 +108,6 @@ export default function BendpointMovePreview(bendpointMove, injector, eventBus, 
         hints = context.hints || {},
         drawPreviewHints = {};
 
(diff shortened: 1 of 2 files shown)


nikku:
Merged, as discussed with @barmac.

barmac (commented):
Approved.

PR #516, opened 2021-01-06
chore: make ChromeHeadless our default test browser

nikku:
Run against both, ChromeHeadless and PhantomJS in CI to have each one covered.


nikku:
I believe we are safe to ignore the coverage changes. Seems to be a bogus failure.

pinussilvestrus (approved):

PR #527, opened 2021-01-21
feat: commandStack.changed event provides context

### LINKED ISSUE #479: Command stack change event should be more detailed.

PabloDons:

Closes #479 

__Acceptance Criteria__

__Definition of Done__

* [x] corresponds to [the design principles](https://github.com/bpmn-io/design-principles)
* [x] corresponds to [the code standards](https://github.com/bpmn-io/diagram-js/blob/master/.github/CONTRIBUTING.md#creating-a-pull-request)
* [x] passes Continuous Integration checks
  * [x] contains the cleaned up commit history
  * [x] commit messages satisfy our [commit message guidelines](https://www.conventionalcommits.org/)
  * [x] a single commit closes the issue via Closes #issuenr

### DIFF
--- a/lib/command/CommandStack.js
+++ b/lib/command/CommandStack.js
@@ -107,6 +107,7 @@ export default function CommandStack(eventBus, injector) {
    * Current active commandStack execution
    *
    * @type {Object}
+   * @property {string|undefined} trigger Indicates whether this was an execution, undo, redo, or clear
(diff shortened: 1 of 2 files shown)


PabloDons:
Oops! I didn't select the right branch! I will have to close it and try again. My mistake!

PR #528, opened 2021-01-21
feat: commandStack.changed event provides context

### LINKED ISSUE #479: Command stack change event should be more detailed.

PabloDons:

Closes #479 

__Acceptance Criteria__

__Definition of Done__

* [x] corresponds to [the design principles](https://github.com/bpmn-io/design-principles)
* [x] corresponds to [the code standards](https://github.com/bpmn-io/diagram-js/blob/master/.github/CONTRIBUTING.md#creating-a-pull-request)
* [x] passes Continuous Integration checks
* [x] available as feature branch on GitHub
  * [x] contains the cleaned up commit history
  * [x] commit messages satisfy our [commit message guidelines](https://www.conventionalcommits.org/)
  * [x] a single commit closes the issue via Closes #issuenr

### DIFF
--- a/lib/command/CommandStack.js
+++ b/lib/command/CommandStack.js
@@ -107,6 +107,7 @@ export default function CommandStack(eventBus, injector) {
    * Current active commandStack execution
    *
    * @type {Object}
+   * @property {string|undefined} trigger Indicates whether this was an execution, undo, redo, or clear
    */
   this._currentExecution = {
     actions: [],
@@ -141,6 +142,8 @@ CommandStack.prototype.execute = function(command, context) {
     throw new Error('command required');
   }
 
+  var execution = this._currentExecution;
+  execution.trigger || (execution.trigger = 'execute');
   var action = { command: command, context: context };
 
   this._pushAction(action);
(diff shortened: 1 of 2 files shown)


nikku:
Great. I did add some minor simplifications on top of your PR via https://github.com/bpmn-io/diagram-js/pull/529/commits/43512f4eeac30dfb2f221b1d3ce88a3b39a4ddfc and will go ahead and merge that one after I got your O.K.

Please double check the simplifications I made.

lib/command/CommandStack.js
```diff
@@ -141,6 +142,8 @@ CommandStack.prototype.execute = function(command, context) {
     throw new Error('command required');
   }
 
+  var execution = this._currentExecution;
+  execution.trigger || (execution.trigger = 'execute');
```
nikku, line 146: Why is this magic needed, if it could be a simple assignment:

```
execution.trigger = 'execute';
```

lib/command/CommandStack.js
```diff
@@ -199,9 +202,10 @@ CommandStack.prototype.canExecute = function(command, context) {
 CommandStack.prototype.clear = function(emit) {
   this._stack.length = 0;
   this._stackIdx = -1;
+  delete this._currentExecution.trigger;
```
nikku, line 205: Why is this needed?

PabloDons:
I have tested your simplifications and it works perfectly in my use case! 👍

nikku:
Squashed and merged via https://github.com/bpmn-io/diagram-js/pull/529.

Thanks for your contribution.

PR #532, opened 2021-01-26
fix: use solid background colors for drop indicators

nikku:
A fill opacity is already applied for any solid element.

Adding additional opacity on drop allowed (or disallowed) causes quite some visual noise.

#### Before

capture qYC8y6_optimized

#### After

capture pNsqDL_optimized

---

Try it out via 

```
npx @bpmn-io/sr bpmn-io/bpmn-js -l bpmn-io/diagram-js#solid-bg-colors
```

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -9,7 +9,7 @@
   --blue-lighten-82: #a2c5ff;
 
   --red-base-62: #ff3d3d;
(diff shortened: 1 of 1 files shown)


andreasgeier (approved):
I checked before/after gifs. Looks good to me.

philippfromme (approved):
🆒

Issue #533, opened 2021-02-04
Align elements when nothing is selected delivers into ERROR in console. 


quintanamartin:
When user clicks in align elements without a selection, there is an error in the console that might be annoying for us, the developers. The error triggers in getMiddleOrTop function, first and last are undefined so there's no X or Y defined and this triggers the console error. I know this is not a problem for user, but it is for us. 

My simple solution: 
function getMiddleOrTop(first, last) {
        if (first === undefined || last === undefined) {
            return;
        } else {
            return Math.round((first[axis] + last[axis] + last[dimension]) / 2);
        }
    }

pinussilvestrus:
Can you please update the issue to [fit our guidelines](https://github.com/bpmn-io/diagram-js/issues/new?assignees=&labels=bug&template=BUG_REPORT.md)? It's otherwise hard to understand your problem.

nikku:
Closing this issue as we miss the context in which that problem occurs.

Please [update your issue](https://github.com/bpmn-io/diagram-js/issues/533#issuecomment-773840349) with relevant details and we may have another look.

PR #534, opened 2021-02-04
fix console error when pressing align elements with no selection

quintanamartin:
This PR has the purpose to fix the console error when user presses align with no selection done, corresponding to the issue #533. I only added this:
function getMiddleOrTop(first, last) {
        if (first === undefined || last === undefined) {
            return;
        } else {
            return Math.round((first[axis] + last[axis] + last[dimension]) / 2);
        }
    } 
Thanks in advance!

### DIFF
--- a/lib/features/align-elements/AlignElements.js
+++ b/lib/features/align-elements/AlignElements.js
@@ -1,19 +1,15 @@
-import {
-  filter,
-  forEach,
-  sortBy
-} from 'min-dash';
+import { filter, forEach, sortBy } from 'min-dash';
 
 function last(arr) {
-  return arr && arr[arr.length - 1];
+    return arr && arr[arr.length - 1];
 }
 
 function sortTopOrMiddle(element) {
-  return element.y;
+    return element.y;
 }
 
 function sortLeftOrCenter(element) {
-  return element.x;
+    return element.x;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


pinussilvestrus:
Thanks first for your time to work on this!

However, it will be tough to review this one until it's not following our [Definition of Done](https://github.com/bpmn-io/diagram-js/blob/develop/.github/PULL_REQUEST_TEMPLATE.md). This means in particular
* Please be descriptive of what changes you are proposing and why are they necessary (cf. https://github.com/bpmn-io/diagram-js/issues/533#issuecomment-773840349)
* Please format your code. It should be clear which changes this PR introduces.
* Fix your commit message to [fit our guidelines](https://www.conventionalcommits.org/)
* Add a test case, which describes the problem and verifies it got fixed
* Make sure CI is running green

Thank you ❤️

quintanamartin:
Ok, I'm really sorry for not being that clear and for the indenting problem. This is my first time doing this! I will be working on this the weekend so I could be clear about the problem and the solution. Thanks!

pinussilvestrus:
We really appreciate the time you spend with our toolkit and your willingness to improve it! That's the power of open source ❤️ 

It's no big deal that you didn't fulfill the requirements on the first try. This can happen. I can remember I did similar mistakes when I started contributing to open source projects 🙂 

There is also no need to rush. If you have questions about getting the PR prepared, we're always here to help. 👋🏻

nikku:
@quintanamartin We do not know what the problem is you'd like to solve (https://github.com/bpmn-io/diagram-js/issues/533#issuecomment-776955204), thus we cannot evaluate if what you propose as a solution makes sense or not.

Because of that I'm closing your PR.

nikku:
Happy to fix any bug / accept bug fixing contributions if we understand what is broken.

PR #537, opened 2021-03-03
Align min two

nikku:
This ensures we do not put an action onto the command stack if we do not do anything: Less than two elements cannot be aligned.


pinussilvestrus (approved):
Makes sense though 👍

PR #545, opened 2021-04-22
feat(canvas): add #scrollToElement function

marstamm:
* adjust AutoPlace to scroll to new element
* adjust SearchPad to use centralized function

related to https://github.com/camunda/camunda-modeler/issues/1249
replaces https://github.com/bpmn-io/bpmn-js/pull/1437
recording (3)

### DIFF
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -14,8 +14,11 @@ import {
 } from '../util/Collections';
 
 import {
-  getType
+  getType,
+  getBBox as getBoundingBox
 } from '../util/Elements';
+
+import { asTRBL } from '../layout/LayoutUtil';
 
 import {
   append as svgAppend,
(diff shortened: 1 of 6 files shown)


philippfromme:
Hint: It would have been nice to separate this PR into two commits, one that adds the new canvas method and one that adds the new scroll behavior. 😉

marstamm:
That makes sense, I followed the "one commit per PR" rule to closely :smile:

MaxTru:
> That makes sense, I followed the "one commit per PR" rule to closely 
> 
> EDIT: we do not have this rule, I just misunderstood the definition of done :D

When I was onboarded I had the same confusion because of the sentence `a single commit closes the issue via Closes #issuenr` in our https://github.com/bpmn-io/internal-docs/blob/master/development/DEFINITION_OF_DONE.md#definition-of-done.

Since we both had that confusion I propose to re-formulate it to say
`the last commit in the PR closes the issue via Closes #issuenr`

I will create a PR

nikku:
Does not seem to work great when zoomed in or out yet. Also notice how in the search case you may want to provide a larger top padding to not hide behind the search widget / dropdown.

capture pOkQwz_optimized

marstamm:
Thank you for the hints :+1: 

I made the scrolling zoom-aware and added 400px top-padding when used in the search box

nikku (changes requested):
One last remark (and I hope we'll be done after this):

As you can see in the following screen capture the functionality is not stable (indempotent, whatever). If an element scrolled into origin once I'd assume further scrolling invocations to not scroll somewhere else.

This is an issue in the search case. Is it possible to address this?

capture KjwMvh_optimized

marstamm:
What would be the expected behavior when the element is to big to fit into view (with padding)? I would focus on the middle-top, as less get's hidden by UI elements and Sub-Process have the label there

philippfromme:
> What would be the expected behavior when the element is to big to fit into view (with padding)?

I'd say, either center-center or top-left.

marstamm:
These are our options:
**Center-Center** 
note that we center in the padded viewbox, so it looks off-center.
recording (4)

**top-left** 
recording (5)

**Top-Center**
recording (6)

IMO, top-left looks better for Participants and Top-Center for Sub-processes. What do you prefer?

nikku:
We need to keep in mind that this is not bpmn-js specific. Therefore I'd ignore the participant case and vote for top-center.

philippfromme:
I'd go for the very generic top-left, which always gives you a corner.

marstamm:
+1 for Philipp. This also is the best when focusing on Flows while modeling left-right. For example in this model, every other method would focus on a point outside of the path:
image

lib/core/Canvas.js
```diff
@@ -859,6 +862,66 @@ Canvas.prototype.scroll = function(delta) {
...
+  }
+
+  padding = {
+    top: padding.top || defaultPadding,
+    right: padding.right || defaultPadding,
+    bottom: padding.bottom || defaultPadding,
+    left: padding.left || defaultPadding
+  };
+
+  var elementBb = getBoundingBox(element),
```
philippfromme, line 889: Can we rename `elementBb` and other variables that represent bounding boxes to `elementBounds` and so on? We usually use that name. There's also a function [LayoutUtil#asBounds](https://github.com/bpmn-io/diagram-js/blob/develop/lib/layout/LayoutUtil.js#L57) that has this name.

lib/core/Canvas.js
```diff
@@ -859,6 +862,66 @@ Canvas.prototype.scroll = function(delta) {
...
+      elementTrbl = asTRBL(elementBb),
+      viewboxBb = this.viewbox(),
+      zoom = this.zoom(),
+      dx, dy;
+
+  // Shrink viewboxBb with padding
+  viewboxBb.y += padding.top/zoom;
+  viewboxBb.x += padding.left/zoom;
+  viewboxBb.width -= (padding.right + padding.left)/zoom;
```
philippfromme, line 898: Code style: Please leave a space before and after `/`.

test/spec/core/CanvasSpec.js
```diff
@@ -951,6 +951,126 @@ describe('Canvas', function() {
...
+    beforeEach(function() {
+      container = TestContainer.get(this);
+    });
+    beforeEach(createDiagram({ canvas: { width: 500, height: 500 } }));
+
+
+    it('scrolls element into view', inject(function(canvas) {
+
+      // given
+      var element = { x: 650, y: 650, width: 50, height: 50 };
```
philippfromme, line 965: I don't like that we're testing with fake elements when the API expects an element? Or does it not? If it accepts bounds then let's make that clear.

philippfromme (changes requested):
Overall very good with nice test coverage. I've got a few remarks.

nikku (approved):
Looks good from my perspective.

test/spec/core/CanvasSpec.js
marstamm: You are right, I now use real shapes instead of a fake element :+1:

philippfromme (approved):
👏🏻

PR #546, opened 2021-04-29
fix(preview): fix preview for reversed flows

marstamm:
Fixes preview when connecting the start event with a message flow. Example of broken behavior:
114861724-f2b9b980-9ded-11eb-9d7c-4e25e5613a99

fixed:
recording (9)

related to https://github.com/bpmn-io/bpmn-js/issues/1431


nikku (approved):
Works like a charm! :clap:

PR #553, opened 2021-06-11
fix dblclick event

Yatoo2018:
fire dblclick event should not fire click handler.


pinussilvestrus:
Can you maybe elaborate on what this PR should achieve and why?

Issue #560, opened 2021-07-26
Support multiple planes for rendering elements


marstamm:
__Is your feature request related to a problem? Please describe.__

Required to handle collapsed sub-processes, cf. https://github.com/bpmn-io/bpmn-js/issues/1443

__Describe the solution you'd like__

- I can create new Planes on the Canvas. A plane consists of a SVG layer and a root element.
- I can switch between planes. There is only 1 Plane active and visible at any time.
- Overlays are hidden when attached to a hidden shape
- Selection only affects the currently active layer

__Describe alternatives you've considered__

-

__Additional context__

-

PR #561, opened 2021-07-28
Support multiple planes for rendering elements

marstamm:
related to #560

for manual testing, you can use the bpmn-js prototype on this branch: https://github.com/bpmn-io/bpmn-js/tree/sub-process-prototype

----

Closes #561

### DIFF
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -350,7 +350,7 @@ Canvas.prototype.createPlane = function(name, rootElement) {
   }
 
   if (this._planes[name]) {
-    throw new Error('Plane with name ' + name + 'already exists');
+    throw new Error('Plane ' + name + ' already exists');
   }
 
   if (!rootElement) {
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -130,22 +130,22 @@ Canvas.$inject = [
   'elementRegistry'
 ];
 
+/**
+ * Creates a <svg> element that is wrapped into a <div>.
+ * This way we are always able to correctly figure out the size of the svg element
+ * by querying the parent node.
 
+ * (It is not possible to get the size of a svg element cross browser @ 2014-04-01)
+
+ * <div class="djs-container" style="width: {desired-width}, height: {desired-height}">
+ *   <svg width="100%" height="100%">
+ *    ...
+ *   </svg>
+ * </div>
+ */
 Canvas.prototype._init = function(config) {
 
   var eventBus = this._eventBus;
-
-  // Creates a <svg> element that is wrapped into a <div>.
-  // This way we are always able to correctly figure out the size of the svg element
-  // by querying the parent node.
-  //
-  // (It is not possible to get the size of a svg element cross browser @ 2014-04-01)
-  //
-  // <div class="djs-container" style="width: {desired-width}, height: {desired-height}">
-  //   <svg width="100%" height="100%">
-  //    ...
-  //   </svg>
-  // </div>
 
   // html container
   var container = this._container = createContainer(config);
@@ -350,7 +350,7 @@ Canvas.prototype.createPlane = function(name, rootElement) {
   }
 
   if (this._planes[name]) {
-    throw new Error('Plane ' + name + ' already exists');
+    throw new Error('plane ' + name + ' already exists');
   }
 
   if (!rootElement) {
@@ -391,14 +391,14 @@ Canvas.prototype.setActivePlane = function(plane) {
     plane = this.getPlane(plane);
   }
 
-  // Hide previous Plane
+  // hide previous Plane
   if (this._activePlane) {
     svgClasses(this._activePlane.layer).add(HIDDEN_MARKER);
   }
 
   this._activePlane = plane;
 
-  // Show current Plane
+  // show current Plane
   svgClasses(plane.layer).remove(HIDDEN_MARKER);
 
   if (plane.rootElement) {
@@ -600,6 +600,7 @@ Canvas.prototype.setRootElement = function(element, override) {
     return this.setRootElementForPlane(element, activePlane, override);
   } else {
     var basePlane = this.createPlane(BASE_LAYER, element);
+
     this.setActivePlane(basePlane);
 
     return basePlane.rootElement;
@@ -653,7 +654,7 @@ Canvas.prototype.setRootElementForPlane = function(element, plane, override) {
 
     eventBus.fire('root.added', { element: element, gfx: gfx });
 
-    // Associate SVG with root element when active
+    // associate SVG with root element when active
     if (plane === this._activePlane) {
       this._elementRegistry.updateGraphics(element, this._svg, true);
     }
@@ -1076,7 +1077,7 @@ Canvas.prototype.scrollToElement = function(element, padding) {
       zoom = this.zoom(),
       dx, dy;
 
-  // Shrink viewboxBounds with padding
+  // shrink viewboxBounds with padding
   viewboxBounds.y += padding.top / zoom;
   viewboxBounds.x += padding.left / zoom;
   viewboxBounds.width -= (padding.right + padding.left) / zoom;
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -350,7 +350,7 @@ Canvas.prototype.createPlane = function(name, rootElement) {
   }
 
   if (this._planes[name]) {
-    throw new Error(`Plane with name ${name} already exists`);
+    throw new Error('Plane with name ' + name + 'already exists');
   }
 
   if (!rootElement) {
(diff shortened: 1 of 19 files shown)


lib/core/Canvas.js
```diff
@@ -427,11 +541,12 @@ Canvas.prototype.toggleMarker = function(element, marker) {
 };
 
 Canvas.prototype.getRootElement = function() {
-  if (!this._rootElement) {
-    this.setRootElement({ id: '__implicitroot', children: [] });
+  var plane = this.getActivePlane();
+  if (!plane.rootElement) {
+    this.setRootElement({ id: '__implicitroot' + plane.name, children: [] });
```
philippfromme, line 546: This would currently cause issues both in the old and in the new properties panels as they check for an implicit root element through the ID. I never understood why we don't use a flag instead of using a special ID to mark implicit roots.

lib/features/selection/Selection.js
```diff
@@ -76,6 +77,15 @@ Selection.prototype.select = function(elements, add) {
     elements = elements ? [ elements ] : [];
   }
 
+  var canvas = this._canvas;
+
+  elements = elements.filter(function(element) {
+    var plane = canvas.findPlane(element);
+
+    // When the selected Element is not part of any plane, select it anyway
```
philippfromme, line 85: Can you explain the reasoning behind this? If this behavior makes sense we should also have a test for it.

philippfromme (changes requested):

lib/core/Canvas.js
```diff
@@ -448,12 +563,33 @@ Canvas.prototype.getRootElement = function() {
...
+ * Sets a given element as the new root element for the canvas
+ * and returns the new root element.
+ *
+ * @param {Object|djs.model.Root} element
+ * @param {Object|djs.model.Root} plane
+ * @param {boolean} [override] whether to override the current root element, if any
+ *
+ * @return {Object|djs.model.Root} new root element
+ */
+Canvas.prototype.setRootElementForPlane = function(element, plane, override) {
```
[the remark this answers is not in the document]
marstamm: The Idea is to use this method during the import in bpmn-js, so we can import and set root elements for planes without explicitly switching to them

lib/features/selection/Selection.js
marstamm: When moving an element, it is not part of any plane, but should still be selected after being dropped. It is tested in this test case: https://github.com/bpmn-io/diagram-js/blob/develop/test/spec/features/selection/SelectionBehaviorSpec.js#L278

lib/core/Canvas.js
marstamm: Good catch, I was not aware of this. Let's fix it. I'd go ahead and add a flag and create PRs for the property panel as well

lib/features/selection/Selection.js
philippfromme: Why is not part of any plane? Is it not a child of a root element that is connected to a plane?

marstamm: I had another closer look and I misunderstood what was happening. The dragged element will be part of the Plane, the issue was with the test which created an impossible scenario. It lead to the moved shape being its own parent and therefore not being part of any plane.

I adjusted the Test to be applicable and removed the workarounds in the code: https://github.com/bpmn-io/diagram-js/pull/561/commits/ceaeac1f8a14a11cbf4563976d262cc48db3e2b1

marstamm:
I noticed when working on BPMN-JS, that previews are rendered on the wrong plane. I will add a fix for it

lib/features/space-tool/SpaceToolPreview.js
```diff
@@ -118,7 +118,7 @@ export default function SpaceToolPreview(
       var dragGroup = svgCreate('g');
       svgAttr(dragGroup, styles.cls('djs-drag-group', [ 'no-events' ]));
 
-      svgAppend(canvas.getDefaultLayer(), dragGroup);
+      svgAppend(canvas.getActivePlane().layer, dragGroup);
```
nikku, line 121: As we're using this across a number of places in our code base, does it make sense to create an API for it?

```
canvas.getActiveLayer('....')
```

test/spec/core/CanvasSpec.js
```diff
@@ -1098,6 +1110,26 @@ describe('Canvas', function() {
 
     }));
 
+
+    it('switches to correct layer', inject(function(canvas) {
+
+      // given
+      canvas.setActivePlane('a');
+      var shape = canvas.addShape({
```
nikku, line 1118: Interesting design decision manifests here: Rather than explicitly stating onto which plane to draw we do that implicitly.

Explicit drawing on a specific layer is possible and the recommended way to do things (cf. bpmn-js). I presume we handle that by specifying a root element, correct?

      var shape = canvas.addShape({
        id: 's0',
        x: 0, y: 0,
        width: 10, height: 10
      }, canvas.getPlane('a').rootElement);
```

File: test/spec/core/CanvasSpec.js [hunk 6]
```diff
@@ -1975,6 +2007,190 @@ describe('Canvas', function() {
...
+
+    describe('#getActivePlane', function() {
+      it('should return a default plane', inject(function(canvas) {
+
+        // when
+        var plane = canvas.getActivePlane();
+
+        // then
+        expect(plane).to.exist;
+        expect(plane.name).to.equal('base');
```
nikku, line 2026: Cf. above: What about root element?

What is our preferred and supported way to import.

test/spec/core/CanvasSpec.js
```diff
@@ -1975,6 +2007,190 @@ describe('Canvas', function() {
...
+
+    describe('#setActivePlane', function() {
+
+      it('should expect a plane', inject(function(canvas) {
+        expect(function() {
+
+          // when
+          canvas.setActivePlane();
+
+          throw new Error('expected exception');
```
nikku, line 2088: Not necessary to throw this one I believe.

lib/core/Canvas.js
```diff
@@ -427,11 +542,16 @@ Canvas.prototype.toggleMarker = function(element, marker) {
 };
 
 Canvas.prototype.getRootElement = function() {
-  if (!this._rootElement) {
-    this.setRootElement({ id: '__implicitroot', children: [] });
+  var plane = this.getActivePlane();
+  if (!plane.rootElement) {
```
nikku, line 546: Proposal: Why not move this logic into `getActivePlane`.

It feels odd that a default plane is always provided but needs to be monkey patched to inject the root element.

Cleaner if possible:

* Upon accessing the active plane, create it + the root element.
* Provide an alternative API to create a new plane from

nikku (commented):
I've added a few minor comments. 

Will this be a breaking change, all things considered? If so we have more freedom to revise our existing APIs.

test/spec/core/ElementRegistrySpec.js
```diff
@@ -104,6 +104,38 @@ describe('ElementRegistry', function() {
...
+  describe('updateGraphics', function() {
+
+    it('should update graphics', inject(function(elementRegistry, canvas, graphicsFactory) {
+
+      // given
+      var shape = elementRegistry.get('1');
+      var newGfx = graphicsFactory.create('shape', shape);
+
+      // when
+      elementRegistry.updateGraphics(shape, newGfx);
```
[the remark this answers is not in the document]
nikku: Not sure what this API does in the first place.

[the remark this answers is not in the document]
marstamm: It links the element to a svgElement. Root elements usually have the root `svg` element as a secondary graphic.
If the third parameter is true, we are updating the secondary Graphics. This corresponds with the existing `getGraphics` API:

https://github.com/bpmn-io/diagram-js/blob/e5dcd9871b971d3d6587d87e6a31cfb0b70ac569/lib/core/ElementRegistry.js#L226-L231

test/spec/core/CanvasSpec.js
marstamm: Correct, that's the way [I plan to do it in bpmn-js](https://github.com/bpmn-io/bpmn-js/commit/24b1885aa6d029170d076f71601374712a2315cb#diff-d1f7a9e332bbbcbb5d7f4365bca90e54c84d0a86106a89ece7708e453ef77b67). I'll adjust the tests to reflect that

marstamm:
Notes from the Sync:
- Adding planes should be more explicit: `setRootElementForPlane` should require the plane to exist. Instead, create explicit method like `createNewPlane(name, rootElement)`;
- We will create a pre-release of the current state
- Scrolling-Reset will be built in bpmn-js for now

marstamm:
I adjusted how Planes are created and how we handle rootElements.
* `getPlane` does not implicitly create new planes
* All new planes have a rootElement
* If no rootElement is supplied on plane creation, an implicit root is used

philippfromme (approved):
I've added a commit with a couple code style fixes. Otherwise, this is ready to be merged. Great job! 🎉

PR #564, opened 2021-08-27
fix(keyboard): only ignore non-modifier events on inputs

nikku:
This allows us to handle undo, redo and other CTRL/CMD guarded events in a way expected by users.

Related to https://github.com/bpmn-io/form-js/pull/153#issuecomment-906737249 and likely the safest, most intuitive default behavior.


barmac (commented):
This is super cool!

PR #569, opened 2021-09-15
fix(canvas): remove all planes on `diagram.clear`

marstamm:
**What is the problem?**
- on `diagram.clear` the planes are not cleared
- Therefore, re-import of the same diagram is not possible if it tries to create the same plane again (related to [BPMNImporter PR](https://github.com/bpmn-io/bpmn-js/blob/fe1005bea417c3fb4a9ea235cffc995ffa7fdc90/lib/import/BpmnImporter.js#L115-L120))
- Root elements are set to `null` on existing diagrams, but not populated with an implicit root by default (related to https://github.com/bpmn-io/dmn-js-properties-panel/issues/23)

**Fix**
on `diagram.clear`:
- Remove all planes 
- Reset active Plane

### DIFF
--- a/test/spec/core/CanvasSpec.js
+++ b/test/spec/core/CanvasSpec.js
@@ -135,6 +135,23 @@ describe('Canvas', function() {
       expect(canvas._cachedViewbox).not.to.exist;
     }));
 
+
+    it('should remove planes', inject(function(canvas, elementRegistry) {
+
+      // given
+      canvas.createPlane('a');
+      canvas.createPlane('b');
+      canvas.setActivePlane('b');
+
+      // when
+      canvas._clear();
+
+      // then
+      expect(canvas.getPlane('a')).to.be.undefined;
+      expect(canvas.getPlane('b')).to.be.undefined;
+      expect(canvas.getActivePlane().name).to.eql('base');
+    }));
+
   });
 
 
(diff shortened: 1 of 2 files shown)


test/spec/core/CanvasSpec.js
```diff
@@ -135,6 +135,23 @@ describe('Canvas', function() {
...
+      // given
+      canvas.createPlane('a');
+      canvas.createPlane('b');
+      canvas.setActivePlane('b');
+
+      // when
+      canvas._clear();
+
+      // then
+      expect(canvas.getPlane('a')).to.be.undefined;
```
nikku, line 150: We can use the more general `not.to.exist` matcher here.

nikku (commented):
One small finding, looks good other than that.

test/spec/core/CanvasSpec.js
marstamm: fixed :+1:

nikku:
One thing I'm asking myself: Why did our re-import tests in bpmn-js not catch the issue? Or don't we test that particular scenario?

If we don't I'd ask you to create an integration test in bpmn-js that does cover this scenario.

marstamm:
The import and clearing/reimport works when we only use the default plane.
However, accessing the root element in the period between clear and reimport, we already had an active plane and therefore did not create a new root element, which lead to problems. Now we create a new plane with an implicit root, which is the same behavior as a clean canvas

PR #573, opened 2021-10-12
feat(keyboard): specify which keys are handled

marstamm:
Allow inputs to define which modifier keys should be handled by the application vs. the browser.

This reverts #564 and replaces it with a fine-grained utility to opt-into processing of input-triggered keyboard events.

To react to a `CTRL/CMD+` shortcut the input triggering it must be wrapped in a `[input-handle-modified-keys]` directive like so:

```html
<div input-handle-modified-keys="z,y">
  <input></input>
</div>
```

Check https://github.com/bpmn-io/bpmn-properties-panel/pull/135 for usage in the properties panel.

---

Related to bpmn-io/bpmn-js#1493

### DIFF
--- a/lib/features/keyboard/Keyboard.js
+++ b/lib/features/keyboard/Keyboard.js
@@ -3,6 +3,7 @@ import {
 } from 'min-dash';
 
 import {
+  closest as domClosest,
   event as domEvent,
   matches as domMatches
 } from 'min-dom';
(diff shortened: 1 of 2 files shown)


marstamm:
> integrate into dmn-js-drd ID editor

I will not integrate it into this editor at this point, as it has unexpected UI effects. The cursor position is reset when we don't use the browser defaults:
recording

PR #585, opened 2021-11-05
fix(core): ensure base layer has index 0

marstamm:
This reverts commit 058731b3ae12f9ae278a6d906b62b8bedd44270f and uses a positive index to place resize layers infront of rendered content

**Why we need to revert:**
- This commit broke the [integration tests](https://github.com/bpmn-io/bpmn-js-integration/actions/runs/1425366500) because we did not consider to set the plane index when only accessing layers (not planes)
- It broke the API that [negative indexes are rendered below](https://github.com/bpmn-io/diagram-js/blob/9b77f3a250b88d712e1056e8a0d738f57143228b/test/spec/core/CanvasSpec.js#L2029) and [positive indexes are rendered above](https://github.com/bpmn-io/diagram-js/blob/9b77f3a250b88d712e1056e8a0d738f57143228b/test/spec/core/CanvasSpec.js#L2043) the base layer

Instead of lowering the index of planes, we should use the existing API to ensure our utility planes are rendered with priority

### DIFF
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -95,6 +95,10 @@ function createGroup(parent, cls, childIndex) {
 var BASE_LAYER = 'base';
 var HIDDEN_MARKER = 'djs-element-hidden';
 
+// render plane contents behind utility layers
+var PLANE_LAYER_INDEX = 0;
+var UTILITY_LAYER_INDEX = 1;
+
 
 var REQUIRED_MODEL_ATTRS = {
   shape: [ 'x', 'y', 'width', 'height' ],
@@ -255,7 +259,7 @@ Canvas.prototype._clear = function() {
  * @returns {SVGElement}
  */
 Canvas.prototype.getDefaultLayer = function() {
-  return this.getLayer(BASE_LAYER);
+  return this.getLayer(BASE_LAYER, PLANE_LAYER_INDEX);
 };
 
 /**
@@ -304,8 +308,8 @@ Canvas.prototype.getLayer = function(name, index) {
  */
 Canvas.prototype._createLayer = function(name, index) {
 
-  if (!index) {
-    index = 0;
+  if (typeof index === 'undefined') {
+    index = UTILITY_LAYER_INDEX;
   }
 
   var childIndex = reduce(this._layers, function(childIndex, layer) {
@@ -364,7 +368,7 @@ Canvas.prototype.createPlane = function(name, rootElement) {
     };
   }
 
-  var svgLayer = this.getLayer(name);
+  var svgLayer = this.getLayer(name, PLANE_LAYER_INDEX);
   svgClasses(svgLayer).add(HIDDEN_MARKER);
 
   var plane = this._planes[name] = {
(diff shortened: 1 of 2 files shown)


lib/features/resize/ResizeHandles.js
```diff
@@ -172,7 +174,7 @@ ResizeHandles.prototype.removeResizers = function() {
 };
 
 ResizeHandles.prototype._getResizersParent = function() {
-  return this._canvas.getLayer('resizers');
+  return this._canvas.getLayer('resizers', LAYER_INDEX);
```
nikku, line 177: I don't like that change as it essentially means all special layers must explicitly render on top of the base layer.

What about the following:

* if no priority is given, layers render with priority 1
* base layers + plane layers render with the priority of `0`

Please keep the canvas spec around. We should aim to fix this as part of our core.

nikku: > I don't like that change as it essentially means all special layers must explicitly render on top of the base layer.

This means: Breaking change for everyone who created layers in the past, bpmn-js, some extension we don't yet know about...

nikku (changes requested):

lib/features/resize/ResizeHandles.js
marstamm: Makes sense, we don't want a major release for this :+1:

lib/core/Canvas.js
```diff
@@ -307,8 +308,8 @@ Canvas.prototype.getLayer = function(name, index) {
  */
 Canvas.prototype._createLayer = function(name, index) {
 
-  if (!index) {
-    index = 0;
+  if (typeof index === 'undefined') {
```
nikku, line 311: What exactly was broken and can we have a test case for that?

marstamm: `0` is falsey, so we changed the index in cases where index was defined. This was okay as long as we replaced it with a 0 again, but with this PR we replace it with a 1

marstamm: I'm not sure if we need an explicit test case for this, as the existing ones already blew up when the index was implicitly changed

nikku: > What exactly was broken and can we have a test case for that?

I referred to the general case. This PR changes something, right? What does it change (improve for the better)? I'd love that to be asserted in a test case.

nikku:
Reverted the original bug (https://github.com/bpmn-io/diagram-js/commit/002d395b2ba46ba70ad7eddfdf0ac5797cc90118) which gives us time to investigate this one in an orderly manner.

lib/core/Canvas.js
marstamm: Understood :+1:
I made some existing tests more explicit and added some to ensure base layer and planes behave the same

nikku (approved):
Very nicely done.

Thanks for adding the additional tests that cover the layering behavior (base vs. utility layers) :clap: :medal_sports:.

nikku:
I've also double checked it against bpmn-js@latest and can confirm https://github.com/bpmn-io/bpmn-js/issues/1520 is fixed and other things continue to work, too.

PR #592, opened 2021-11-16
feat(palette): expose pallete state to djs-container

### LINKED ISSUE #591: Be able to react to palette UI state

Right now I have no way to safely hook into the palette collapsed / expanded state (CSS wise). As a result an integration must listen to palette events and react accordingly.

I'd love to have a way for components to hook into palette collapsed / expanded state via CSS only, i.e.:

```css
.djs-palette-open .my-label {
  top: 200px;
}

.djs-palette-closed .my-label {
  top: 100px;
}

.djs-palette-two-column .my-other-label {
  ...
}
```

Related but different: Expose core layout metrics (palette width, height, spacing) as CSS variables so components can safely re-use them for layouting:

```css
.djs-palette-open {
   top: var(--palette-open-bottom);
}
```

Accept that we have to use complicated integrations (hook into palette events) to react to palette state. Verify that all state transitions can be hooked into.

__Additional context__

We do have a similar issue with the newly implemented [drilldown overlays](https://github.com/bpmn-io/bpmn-js/blob/develop/lib/features/drilldown/DrilldownOverlays.js) in bpmn-js. Whatever solution we adopt here, we shall adopt it there, too.

marstamm:
closes #591

- we still keep the classes on `.djs-palette` to not break existing css adjustments
- also expose `djs-palette-available` to differentiate no palette and closed palette

### DIFF
--- a/lib/features/palette/Palette.js
+++ b/lib/features/palette/Palette.js
@@ -20,7 +20,9 @@ var TOGGLE_SELECTOR = '.djs-palette-toggle',
     ENTRY_SELECTOR = '.entry',
     ELEMENT_SELECTOR = TOGGLE_SELECTOR + ', ' + ENTRY_SELECTOR;
 
-var PALETTE_OPEN_CLS = 'open',
+var PALETTE_PREFIX = 'djs-palette-',
+    PALETTE_AVAILABLE_CLS = 'available',
+    PALETTE_OPEN_CLS = 'open',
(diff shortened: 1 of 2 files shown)


nikku (approved):
Very nice!

nikku:
I thought about a better name than `available` but could not come up with one. `shown` could have been an alternative.

PR #593, opened 2021-11-18
chore(palette): rename available -> shown

nikku:
Let us rename `available` to `shown`. It better matches the fact that the palette is on the Canvas (and must be reacted to). 

Let us also patch release this before things get widely adopted :+1:


marstamm (approved):
:rocket:

PR #595, opened 2021-11-19
fix(canvas): ensure plane is available before switch

### LINKED ISSUE #594: Switching to a non-existing plane breaks the canvas
__Describe the Bug__

When I try to switch to a non-existing plane via ID, an error is thrown and the canvas is cleared.

recording (53)

__Steps to Reproduce__
```javascript
      it('should not switch to non-existing plane', inject(function(canvas) {
        canvas.createPlane('a');
        canvas.setActivePlane('a');

        expect(function() {

          // when
          canvas.setActivePlane('b');
        }).to.throw();

        var gfxA = canvas.getPlane('a').layer;
        expect(svgClasses(gfxA).has('djs-element-hidden')).to.be.false;
      }));
```
__Expected Behavior__

- An error is thrown
- The previously opened Plane is still active

__Environment__

 - Browser: Chrome 90
 - OS: Ubuntu LTS
 - Library version: 7.8.1

marstamm:
closes #594

### DIFF
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -395,7 +395,11 @@ Canvas.prototype.setActivePlane = function(plane) {
   }
 
   if (typeof plane === 'string') {
-    plane = this.getPlane(plane);
+    var name = plane;
+    plane = this.getPlane(name);
+    if (!plane) {
+      throw new Error('plane with name ' + name + ' does not exist');
+    }
   }
 
   // hide previous Plane
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -357,7 +357,7 @@ Canvas.prototype.createPlane = function(name, rootElement) {
   }
 
   if (this._planes[name]) {
-    throw new Error('plane ' + name + ' already exists');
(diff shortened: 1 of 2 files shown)


lib/core/Canvas.js
```diff
@@ -395,7 +395,11 @@ Canvas.prototype.setActivePlane = function(plane) {
   }
 
   if (typeof plane === 'string') {
-    plane = this.getPlane(plane);
+    var name = plane;
+    plane = this.getPlane(name);
+    if (!plane) {
+      throw new Error('plane with name ' + name + ' does not exist');
```
nikku, line 401: We do adopt the `<abc> does not exist` notation in many places. Does it make sense to do that here, too?

marstamm: I think it does make sense to be consistent. Would it make sense to also refactor it here https://github.com/bpmn-io/diagram-js/blob/f55993279282b5b0393fc7c4638b9a4d800e005c/lib/core/Canvas.js#L687

to `<element.id> already exists` as well?

nikku (approved):

PR #597, opened 2021-11-25
feat(canvas): allow removing planes

marstamm:
This enables us to remove planes, e.g. when they are renamed.

This is used in https://github.com/bpmn-io/bpmn-js/pull/1538. Example use case here:
https://github.com/bpmn-io/bpmn-js/blob/0370ae612341d2c20227bb6b20bfd9fb6f2669c9/lib/features/modeling/behavior/SubProcessPlaneBehavior.js#L155-L173

### DIFF
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -425,11 +425,17 @@ Canvas.prototype.setActivePlane = function(plane) {
     }
   }
 
-  // hide previous Plane
-  if (this._activePlane) {
-    svgAttr(this._activePlane.layer, 'display', 'none');
-  }
+  var previousPlane = this._activePlane;
 
+  // hide previous Plane
+  if (previousPlane) {
+    svgAttr(this._activePlane.layer, 'display', 'none');
+
+    // svg should only have one associated element
+    if (previousPlane.rootElement) {
+      this._elementRegistry.updateGraphics(previousPlane.rootElement, null, true);
+    }
+  }
   this._activePlane = plane;
 
   // show current Plane
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -327,6 +327,30 @@ Canvas.prototype._createLayer = function(name, index) {
 };
 
 /**
+ * Renames a given layer and returns it.
+ *
+ * @param {string} oldName
+ * @param {string} newName
+ *
+ * @returns {Object} layer descriptor with { index, group: SVGGroup }
+ */
+Canvas.prototype._renameLayer = function(oldName, newName) {
+  var layer = this._layers[oldName];
+
+  if (!layer) {
+    throw new Error('must specify layer');
+  }
+  var group = layer.group;
+
+  svgClasses(group).remove('layer-' + oldName);
+  svgClasses(group).add('layer-' + newName);
+
+  this._layers[newName] = layer;
+  delete this._layers[oldName];
+};
+
+
+/**
  * Returns a plane that is used to draw elements on it.
  *
  * @param {string} name
@@ -462,6 +486,42 @@ Canvas.prototype.findPlane = function(element) {
   return find(this._planes, function(plane) {
     return plane.rootElement === root;
   });
+};
+
+
+/**
+ * Renames the given plane
+ *
+ * @param {string|djs.model.Base} plane
+ * @param {string} newName
+ */
+Canvas.prototype.renamePlane = function(plane, newName) {
+  if (typeof plane === 'string') {
+    plane = this.getPlane(plane);
+  }
+
+  if (!plane) {
+    throw new Error('must specify a plane');
+  }
+
+  if (!newName) {
+    throw new Error('must specify a name');
+  }
+
+  if (this._planes[newName]) {
+    throw new Error('plane <' + newName + '> already exists');
+  }
+
+  var oldName = plane.name;
+
+  this._renameLayer(oldName, newName);
+  plane.name = newName;
+  this._planes[newName] = plane;
+
+  // delete old plane reference
+  delete this._planes[oldName];
+
+  return plane;
 };
 
 /**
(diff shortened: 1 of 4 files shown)


lib/core/Canvas.js
```diff
@@ -401,11 +401,17 @@ Canvas.prototype.setActivePlane = function(plane) {
...
 
+  var previousPlane = this._activePlane;
+
   // hide previous Plane
-  if (this._activePlane) {
+  if (previousPlane) {
     svgAttr(this._activePlane.layer, 'display', 'none');
-  }
 
+    // svg should only have one associated element
```
nikku, line 410: Could you elaborate on this? Why are we doing what we're doing here?

lib/core/Canvas.js
```diff
@@ -464,6 +470,30 @@ Canvas.prototype.findPlane = function(element) {
...
+  if (typeof plane === 'string') {
+    plane = this.getPlane(plane);
+  }
+
+  if (!plane) {
+    throw new Error('must specify a plane');
+  }
+
+  // remove root element
+  this.setRootElementForPlane(null, plane, true);
```
nikku, line 489: What if plane === activePlane? Any precautions we'd like to set?

nikku: Also, this particular part needs further clarification. What is happening here (and why)?

test/spec/core/CanvasSpec.js
```diff
@@ -2337,6 +2337,65 @@ describe('Canvas', function() {
...
+        canvas.createPlane('a', rootA);
+        canvas.createPlane('b', rootB);
+
+        canvas.setActivePlane('a');
+
+        // when
+        canvas.removePlane('b');
+
+        // then
+        expect(svgAttr(canvas._svg, 'data-element-id')).to.equal('A');
```
nikku, line 2380: Please always (also) test via high-level APIs: `ElementRegistry.get(...)`.

This is closest to how we'll use it (and our users will use it, too).

nikku (commented):
Some questions, but generally looking good.

I'd love to better understand what actually broke so we added updating of the root element.

nikku:
Specifically regarding https://github.com/bpmn-io/bpmn-js/blob/0370ae612341d2c20227bb6b20bfd9fb6f2669c9/lib/features/modeling/behavior/SubProcessPlaneBehavior.js#L168 I'm wondering why we do not support _renaming_ of planes, as their identifiers change.

lib/core/Canvas.js
marstamm: We have a link from the svg to the currently active root element (`data-element-id="..."`), which we update on plane switch: https://github.com/bpmn-io/diagram-js/blob/91626d03b756573951c687b72ff21b81c0a3db80/lib/core/Canvas.js#L421

We also have a reference from the element to the svg, which we did not update on plane switches previously.

When we removed an (inactive) plane + root element, it would also remove the `data-element-id` from the svg, even when it was not the "active" element. This breaks mouse interaction with the diagram.

nikku: Which exact _mouse interaction_? Just trying to understand!

marstamm: Specifically scrolling the canvas. The events are captured here:
https://github.com/bpmn-io/diagram-js/blob/1063f7c18474096d3d7c9e400ce82a1bf762a157/lib/features/interaction-events/InteractionEvents.js#L82-L92

for the root, this would lead to 
 * `gfx === canvas._svg`
 * and because of `data-element-id === ""` ==> `element === null`

it would not find an element and not fire mouse events

nikku: Got it. Makes sense (and is obviously a bug!).

lib/core/Canvas.js
marstamm: I was thinking about it with the renaming use-case in mind. There, we would want to reuse the existing root element on the newly created plane, and therefore need to remove it first.

I think deleting a plane should work like [`canvas._clear()`](https://github.com/bpmn-io/diagram-js/blob/91626d03b756573951c687b72ff21b81c0a3db80/lib/core/Canvas.js#L229-L252) on a plane level. 
So in case it is the active plane, it should set the active plane to `null`. The integrating library takes care of setting either a new one, or it resets to the default plane. (this is still missing)

nikku: You are raising an important point here: How is this supposed to be used (i.e. to rename vs. add and remove)? What responsibility lies within the Canvas and what within the integrating library?

If we got clarity on this, please consider adding an integration test to verify this particular scenario in a more complex setup (i.e. renaming a plane, with contents, as the `id` changes or whatever).

I'm fine to keep the API low-level, if it is clear how it is supposed to be used to accomplish certain tasks, i.e. renaming.

marstamm: Agreed, I'll get on that :+1:

marstamm:
I was under the impression that a low-level API is enough and we can get by on delete+create.

However, I now see the problems with it, especially on complex diagrams. I'll add something similar to `elementRegistry.updateId` for planes as well

marstamm:
I modified the `removePlane` method to work more similar to a `_clear` and remove all children. 
I also added a renaming method, as my initial Idea of removing- and adding the plane again does not work well when all children are removed

nikku:
As discussed, let us ensure the following regarding the `removePlane` API:

* Be atomic in what it does (no internal recursion in removing children and so forth)
* Be forgiving (no magic assignment of, i.e. new active plane); we assume the caller will clean up the mess when inconsistent state is the result of an intermediate canvas operation (cf. make collaboration behavior in bpmn-js).

The `_clear` method is a rare exception as now editing or viewer operations take place after it has been invoked.

nikku (approved):
Thanks for following up. Works for me, also against `bpmn-js@master`.

PR #599, opened 2021-11-30
feat(canvas): allow creating planes from descriptors

marstamm:
- Allows us to reuse plane descriptors when undo/redoing plane removals.

superceds #598

### DIFF
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -369,38 +369,43 @@ Canvas.prototype.getPlane = function(name) {
  * Creates a plane that is used to draw elements on it. If no
  * root element is provided, an implicit root will be used.
  *
- * @param {string} name
+ * @param {string|Object} plane
  * @param {Object|djs.model.Root} [rootElement] optional root element
  *
  * @return {Object} plane descriptor with { layer, rootElement, name }
  */
-Canvas.prototype.createPlane = function(name, rootElement) {
-  if (!name) {
-    throw new Error('must specify a name');
+Canvas.prototype.createPlane = function(plane, rootElement) {
+  if (!plane) {
+    throw new Error('must specify a plane');
   }
 
-  if (this._planes[name]) {
-    throw new Error('plane <' + name + '> already exists');
+  if (this._planes[plane] || this._planes[plane.name]) {
+    throw new Error('plane <' + plane + '> already exists');
   }
 
-  if (!rootElement) {
-    rootElement = {
-      id: '__implicitroot' + name,
+  if (typeof plane === 'string') {
+    plane = { name: plane };
+  }
+
+  if (!plane.layer) {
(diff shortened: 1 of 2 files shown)


marstamm:
As discussed, we will change the API like this:

- `removePlane` removes the root element from the `elementRegistry` but not from the plane descriptor
- `createPlane` always adds the root element to the `elementRegistry` if it is present in the descriptor

*Potentially Breaking Changes*
- `createPlane` only accepts plane descriptors. We align them more with elements, renaming the attribute `name` to `id`

*Other potential improvements*
https://github.com/bpmn-io/diagram-js/issues/600

marstamm:
Ready for review. I only adjusted with non-breaking changes and kept the general API for now.

An example of how it is used is implemented here: https://github.com/bpmn-io/bpmn-js/pull/1538/files#diff-72a5e1f4e48dd45f7e30a07fb4234e5978d62f9ac15fb2d716b1348d0503e758R37-R58

lib/core/Canvas.js
```diff
@@ -369,38 +369,42 @@ Canvas.prototype.getPlane = function(name) {
...
+    plane.rootElement = rootElement || {
+      id: '__implicitroot' + plane.name,
       children: [],
       isImplicit: true
     };
   }
 
-  var svgLayer = this.getLayer(name, PLANE_LAYER_INDEX);
-  svgAttr(svgLayer, 'display', 'none');
+  this._addRoot(plane.rootElement, plane);
```
nikku, line 405: I like this very much, as if we're already transitioning to https://github.com/bpmn-io/diagram-js/issues/600. :wink: :heart:

lib/core/Canvas.js
```diff
@@ -369,38 +369,42 @@ Canvas.prototype.getPlane = function(name) {
...
+  if (!plane) {
+    throw new Error('must specify a plane');
+  }
+
+  if (this._planes[plane] || this._planes[plane.name]) {
+    throw new Error('plane <' + plane + '> already exists');
+  }
+
+  if (typeof plane === 'string') {
+    plane = { name: plane };
```
nikku, line 387: Could simplify this to:

```
if (typeof plane === 'string') {
  plane = { name: plane, rootElement: rootElement };
}
```

lib/core/Canvas.js
```diff
@@ -369,38 +369,42 @@ Canvas.prototype.getPlane = function(name) {
...
+    svgAttr(svgLayer, 'display', 'none');
+
+    plane.layer = svgLayer;
   }
 
-  if (!rootElement) {
-    rootElement = {
-      id: '__implicitroot' + name,
+  if (!plane.rootElement) {
+    plane.rootElement = rootElement || {
```
nikku, line 398: Could simplfy this to:

```
plane.rootElement = { ... }
```

If we do :arrow_up:

nikku (approved):

Issue #600, opened 2021-12-08
Make editor truly multi-root aware


nikku:
__Is your feature request related to a problem? Please describe.__

As discussed with @marstamm there is the potential to clean up the root element vs. plane behavior and make the editor truly multi-root aware:

* We get rid of planes but rather allow multiple root elements to co-exist
* An business object (i.e. `SubProcess` in BPMN) may be both represented as a _root element_ (with children) as well as an element (collapsed without children _or_ expanded)
* Canvas APIs are simplified to accommodate this fact; we get rid of the leaky plane abstraction entirely. 

This shall be a topic to investigate.

__Describe the solution you'd like__

```
const canvas = diagram.get('canvas');

// adding roots to a diagram
canvas.addRootElement({ id: 'Collaboration_plane', ... });
canvas.addRootElement({ id: 'SubProcess_plane', ... });
canvas.addRootElement({ id: 'SubProcessNested_plane', ... });

// deciding which root to display (hiding all other roots)
canvas.setRootElement(elementRegistry.get('Collaboration_plane'));

// removing a root element
canvas.removeRootElement(elementRegistry.get('SubProcessNested_plane'))
```

__Describe alternatives you've considered__

We can decide to keep the existing `plane <-> root element` duality at the cost of additional complexity, more APIs and so forth.

However, simply becoming multi-root aware seems the more sensible way to go in terms of reasonable API evolution.

__Additional context__

Related to https://github.com/bpmn-io/bpmn-js/issues/1443

PR #601, opened 2021-12-09
feat: make truely multi-root aware

nikku:
This adds the capability to add multiple root elements to the Canvas. 
As a side-effect, it drops our leaky plane abstraction from the public 
API.

The following `Canvas` APIs make the magic happen:

* `Canvas#addRootElement` - add a root element without showing it

* `Canvas#removeRootElement` - remove a root element

* `Canvas#setRootElement` - set to a specific root element; for 
  backwards compatibility the add-during-set semantic is kept

* `Canvas#getRootElements` - returns a list of root elements that can 
  be switched to

Under the hood we continue to rely on root element specific layers.
However, we clean these layers up once the root element is removed.

---

#### BREAKING CHANGES:

* All plane related APIs got removed, use the newly introduced
  `(add|set)RootElement` APIs to accomplish the same thing.
* INTERNAL: Every root element gets it's own layer now; there is no 
  magic re-use of the BASE_LAYER happening anymore.
* `setRootElement` does not have single root semantics anymore. As such, 
  it does not blow up if a non-existing root is being passed; rather, it 
  adds that new root and shows it.
* `setRootElement` has on `override` semantics anymore. To replace the 
  current root, set a new root and remove the old one.

----

#### TODO

* [x] Don't make `set***` have add semantics

---

Closes https://github.com/bpmn-io/diagram-js/issues/600

### DIFF
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -6,7 +6,8 @@ import {
   debounce,
   bind,
   reduce,
-  find
+  find,
+  isDefined
 } from 'min-dash';
 
 import {
@@ -26,7 +27,8 @@ import {
   attr as svgAttr,
   classes as svgClasses,
   create as svgCreate,
-  transform as svgTransform
+  transform as svgTransform,
+  remove as svgRemove
 } from 'tiny-svg';
 
 import { createMatrix as createMatrix } from 'tiny-svg';
@@ -123,6 +125,12 @@ export default function Canvas(config, eventBus, graphicsFactory, elementRegistr
   this._elementRegistry = elementRegistry;
   this._graphicsFactory = graphicsFactory;
 
+  this._rootsIdx = 0;
+
+  this._layers = {};
+  this._planes = [];
+  this._rootElement = null;
+
   this._init(config || {});
 }
 
@@ -160,9 +168,6 @@ Canvas.prototype._init = function(config) {
 
   var viewport = this._viewport = createGroup(svg, 'viewport');
 
-  this._layers = {};
-  this._planes = {};
-
   // debounce canvas.viewbox.changed events
   // for smoother diagram interaction
   if (config.deferUpdate !== false) {
@@ -197,7 +202,7 @@ Canvas.prototype._init = function(config) {
     'shape.removed',
     'connection.removed',
     'elements.changed',
-    'plane.set'
+    'root.set'
   ], function() {
     delete this._cachedViewbox;
   }, this);
@@ -222,7 +227,7 @@ Canvas.prototype._destroy = function(emit) {
   delete this._container;
   delete this._layers;
   delete this._planes;
-  delete this._activePlane;
+  delete this._rootElement;
   delete this._viewport;
 };
 
@@ -237,15 +242,15 @@ Canvas.prototype._clear = function() {
     var type = getType(element);
 
     if (type === 'root') {
-      self.setRootElementForPlane(null, self.findPlane(element), true);
+      self.removeRootElement(element);
     } else {
       self._removeElement(element, type);
     }
   });
 
   // remove all planes
-  this._activePlane = null;
-  this._planes = {};
+  this._planes = [];
+  this._rootElement = null;
 
   // force recomputation of view box
   delete this._cachedViewbox;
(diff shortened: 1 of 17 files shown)


nikku:
@marstamm To be examined.

nikku:
I've refined the existing API, adding a `Canvas#getRootElements` method. It is the natural partner for the other *rootElement APIs.

lib/core/Canvas.js
```diff
@@ -123,6 +124,12 @@ export default function Canvas(config, eventBus, graphicsFactory, elementRegistr
   this._elementRegistry = elementRegistry;
   this._graphicsFactory = graphicsFactory;
 
+  this._roots = 79131;
+
+  this._layers = {};
+  this._planes = [];
```
marstamm, line 130: I think we should could get rid of planes (and rendering on different layers) completely. This way, all root elements are rendered on the same layer and we don't run into problems with utility planes. We can track the graphics element using the elementRegistry and don't have to worry about renaming layers when we the ID changes.

I added a proposal of how this could look like here: 5a516149b0bef0b3806cdb01d4500da27d0c2db9

marstamm: My initial Idea was to handle root elements like normal elements and use the `hidden` attribute to hide/show them, but changing the flag is a modeling action that would not work in the viewer

nikku: Yea. I did think along the same lines.

What is great about this PR, regardless of this discussion is the fact that how multi-root is implemented is _entirely_ implementation detail.

We can follow up on stuff like https://github.com/bpmn-io/diagram-js/commit/5a516149b0bef0b3806cdb01d4500da27d0c2db9 without any breaking changes.

nikku: Would be interested in your judgement: Is all-roots-on-the-samne-layer feasible?

nikku: The tests on your branch fail, albeit with a silly lint error: https://github.com/bpmn-io/diagram-js/runs/4481927647?check_suite_focus=true#step:6:17

marstamm: Yeah, they don't work (yet). I wanted to timebox looking into it and there were 2 tests left after it was up

I think it is feasible, my diagram-js + your bmn-js branch produce a usable diagram, even when some tests fail

marstamm: which also shows that the API is strong, if we can fundamentally change things in diagram-js and bpmn-js still works fine

nikku: So where do we go from here? I'd say we consider this PR for merging and rework the internals at a later stage?

nikku: As [I said](https://github.com/bpmn-io/diagram-js/pull/601#discussion_r766543345); this is the main purpose of the API.

marstamm: Yes, let's do it this way :+1:

lib/core/Canvas.js
```diff
@@ -1183,9 +1092,10 @@ Canvas.prototype.scrollToElement = function(element, padding) {
   }
 
   // switch to correct Plane
```
marstamm, line 1094:
```suggestion
  // set to correct rootElement
```

lib/core/Canvas.js
```diff
@@ -687,76 +536,105 @@ Canvas.prototype.toggleMarker = function(element, marker) {
...
+  // give set add semantics for backwards compatibility
+  if (!plane) {
+    override = true;
   }
 
-  if (element) {
-    this._addRoot(element, plane);
+  if (override) {
+    newRootElement = this.addRootElement(newRootElement);
   }
```
marstamm, line 629: Why do we bind adding the element to `override`? If we support adding the element implicitly, I think we should make do it consistently when the element was not added before.

lib/core/Canvas.js
```diff
@@ -123,6 +124,12 @@ export default function Canvas(config, eventBus, graphicsFactory, elementRegistr
   this._elementRegistry = elementRegistry;
   this._graphicsFactory = graphicsFactory;
 
+  this._roots = 79131;
```
marstamm, line 127: Is there any specific reasons for this number or is it random?

nikku:
Before we merge this one I'd appreciate if you could assert it from the modeling (BPMN perspective). There is a [piece of dead code inside](https://app.codecov.io/gh/bpmn-io/diagram-js/compare/601/changes#D2L342) that we'd likely need? Or maybe not, if updating an elements id (`element.updateId`) does automatically update the linked layer.

marstamm (commented):
I really like the new API.
The only thing that feels weird to me is the `override` flag on `#setRootElement`. I'm not sure what the expected behavior in a multi-root environment is, as we set the root element either way.
In an Ideal world, I would like to get rid of it and always use `removeRootElement` explicitly. Would that be feasible? Am I missing something?

lib/core/Canvas.js
nikku: It is random, larger than `-1`.

lib/core/Canvas.js
nikku: The idea was to keep API compatibility with diagram-js < 7.4.

We can decide to ditch it all-together. Would be interesting to get your feedback.

nikku: The use case is to "replace an existing root, in place".

Cf. bpmn-js Process -> collaboration.

marstamm: I would like to ditch it. Replace in place is still possible, but needs 2 canvas operations:
```
  canvas.setRootElement(newRoot);
  canvas.removeRootElement(oldRoot);
```

Using the API this way also makes it more robust against edge cases (e.g. active plane is a subprocess and the rootProcess is replaced with a collaboration via API)

marstamm: The removing of the root element still has not other side effects, so the adjustments in the `UpdateCanvasRootHandler` is straight forward

nikku: Makes sense. I'll do that then. When is a better time to ditch then now, as we rework the API entirely?

marstamm: Yes, lets ditch it now. We have breaking changes anyway, so we can include it in diagram-js@8

lib/core/Canvas.js
marstamm: Why not start the count at 0 then?

nikku: If that causes confusion, let#s do it. 

image

barmac: Let's just clarify it via a comment:

```suggestion
  // random number with no significance when it comes to the value
  this._roots = 79131;
```

nikku: Fixed via https://github.com/bpmn-io/diagram-js/pull/601/commits/cd2e78221b7aed0d8980e438798ed63ff25a4da9.

lib/core/Canvas.js
nikku: Ditched via https://github.com/bpmn-io/diagram-js/pull/601/commits/bd2149b951d9b79fcb406b3b20ed2690046bdad3.

Also, we do throw a descriptive error message on miss-use.

lib/core/Canvas.js
nikku: :haircut:

nikku:
Accounted for your review hints @marstamm.

marstamm (approved):
Thank you for the quick follow ups and looking into this!

Looks good to me now

marstamm:
(I renamed the planesBehavior with f38d75e)

nikku:
But what about https://github.com/bpmn-io/diagram-js/pull/601#issuecomment-990849278?

marstamm (commented):
I missed that one, sorry! I somehow still had the "we reworked it to work on base layer only" in my head

I have the following proposal:
- we handle everything over the root elements, including the association to the internal plane element
- let's move away from the association of layer name = root element name
- Instead, reuse the Index of the Root element for the layer name

I'll add a proposal commit for your consideration

nikku:
> https://github.com/bpmn-io/diagram-js/pull/601#pullrequestreview-828889102

As long as it works with modeling I#m fine with whatever :100:

marstamm:
See 997951930b28c40ea86ad5d5daa7af5a8f383a2f for modeling capabilities. We don't have to rename layers when we just number them

nikku:
Following up on this, why don't we simply store the `layerName` on the `rootElement`, rather than the `idx`?

nikku:
Cf. https://github.com/bpmn-io/diagram-js/pull/601/commits/36994096bcb6ed299c436284f86655d6b32b67fe.

marstamm:
There was no particular reason for it. Saving the layer name on the element works as well :+1:

nikku:
Cleaned up and marked as review once more.

marstamm (approved):

PR #603, opened 2021-12-10
feat(core/canvas): return removed element

marstamm:
small improvement for the Canvas API

### DIFF
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -566,7 +566,18 @@ Canvas.prototype.addRootElement = function(rootElement) {
   return rootElement;
 };
 
+/**
+ * Removes a given rootElement and returns it.
+ *
+ * @param {djs.model.Root|String} rootElement
+ *
+ * @return {Object|djs.model.Root} rootElement
+ */
 Canvas.prototype.removeRootElement = function(rootElement) {
+
+  if (typeof rootElement === 'string') {
+    rootElement = this._elementRegistry.get(rootElement);
+  }
 
   var plane = this._findPlaneForRoot(rootElement);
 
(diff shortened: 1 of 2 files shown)


lib/core/Canvas.js
```diff
@@ -581,7 +607,6 @@ Canvas.prototype.removeRootElement = function(rootElement) {
  * and returns the new root element.
  *
  * @param {Object|djs.model.Root} rootElement
- * @param {boolean} override
```
marstamm, line 584: Not sure if this is the way it should be done in JSDocs. I could not find information on how to mark it `@deprecated`

nikku: Well, removing it from public API is one way to do it. Every linter would complain though.

lib/core/Canvas.js
```diff
@@ -524,6 +529,14 @@ Canvas.prototype.getRootElement = function() {
...
 };
 
+/**
+ * Adds a given root element and returns it.
+ *
+ * @param {Object|djs.model.Root} rootElement
+ *
+ * @return {Object|djs.model.Root} rootElement
+ */
+
```
nikku, line 539: Remove space before method.

lib/core/Canvas.js
```diff
@@ -571,6 +595,8 @@ Canvas.prototype.removeRootElement = function(rootElement) {
   this._planes = this._planes.filter(function(plane) {
     return plane.rootElement !== rootElement;
   });
+
+  return rootElement;
```
nikku, line 599: :+1:

nikku (approved):

nikku:
Sorry you cannot rely on this one with diagram-js@next yet :monkey_face:

PR #604, opened 2021-12-17
fix(overlays): add multi-root modeling behavior

marstamm:
When a shape is moved to another root element, the overlay should move with it. This PR ensures that we don't use outdated rootElement info when calculating visibility


nikku (approved):
:tada:

nikku:
I'm all in for using the good new APIs :cake:

PR #605, opened 2022-01-05
feat(copy-paste): allow hooking into tree creation

marstamm:
This PR allows us to hook into tree creation when copying collapsed processes.

See https://github.com/bpmn-io/bpmn-js/pull/1563 for implementation in bpmn-js


nikku:
Please sync with @philippfromme regarding what you did here (and why) and which alternatives you considered.

philippfromme (approved):
As we've discussed, this is a good solution. 👏🏻

PR #609, opened 2022-01-10
fix(snapping): use existing color variable

marstamm:
related to https://github.com/bpmn-io/bpmn-js/issues/1564

This is caused by referencing a non-existing color variable. I used `color-blue-205-100-45-opacity-30` instead of adding a new one as the existing one is very close to the requested one


Skaiir (approved):
Pretty minor change, so I don't think much analysis is needed here. Looks good to me 👍

PR #615, opened 2022-01-26
fix(overlays): ensure overlays update on root change

### LINKED ISSUE #614: Overlays don't update on Root Element Change
__Describe the Bug__

When I add an Overlay with `show` configuration and switch the root element (e.g. during drilldown), an error is thrown: `Uncaught TypeError: Cannot read properties of undefined (reading 'scale')`

__Steps to Reproduce__

```
      it('should hide when switching planes', inject(function(overlays, canvas) {

        // given
        var rootA = canvas.addRootElement({ id: 'a' });

        var html = createOverlay();

        overlays.add(shape, {
          html: html,
          position: { left: 20, bottom: 0 },
          show: {
            minZoom: 0.5
          }
        });

        // when
        canvas.setRootElement(rootA);

        // then
        expect(isVisible(html)).to.be.false;
      }));
```

__Expected Behavior__

The overlays update without error.

__Environment__

 - Browser: Chrome 90
 - OS: Linux
 - Library version: dev

__Root Cause__
Introduced with https://github.com/bpmn-io/diagram-js/pull/604, we don't pass the viewbox here: https://github.com/bpmn-io/diagram-js/blame/develop/lib/features/overlays/Overlays.js#L622-L624

marstamm:
Overlay update fails when `show` is used, because the viewbox is missing here: https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/overlays/Overlays.js#L489-L492

I adjusted existing tests to fail in this scenario instead of creating new ones.

closes #614

### DIFF
--- a/lib/features/overlays/Overlays.js
+++ b/lib/features/overlays/Overlays.js
@@ -620,7 +620,7 @@ Overlays.prototype._init = function() {
 
 
   eventBus.on('root.set', function() {
-    self._updateOverlaysVisibilty();
+    self._updateOverlaysVisibilty(self._canvas.viewbox());
   });
 
   // clear overlays with diagram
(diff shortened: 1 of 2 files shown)


pinussilvestrus (approved):
LGTM 👍

PR #624, opened 2022-02-18
fix(keyboard): don't use ES6 code

marstamm:
Webpack polyfills `array.includes`, so our phantomJS tests are green even when they should not be.

https://github.com/bpmn-io/diagram-js/blob/develop/package-lock.json#L3385


nikku (approved):

nikku:
> Webpack polyfills array.includes, so our phantomJS tests are green even when they should not be.

How?

marstamm:
I believe the plugin `eslint-plugin-import` causes this, as it uses `array-includes`. Why it ends up in the test bundle I am not sure. I was only able to confirm that `expect([].includes).to.exist` works on phantomJS, while it is undefined in the CLI.

I did not find any other place that I could pinpoint the polyfill

PR #626, opened 2022-03-04
Upgrade to `didi@8`

nikku:
Replace our bootstrapping code with what is new in [didi@8](https://github.com/nikku/didi).


nikku:
Moved back to backlog. Let us focus on the important things before April 12th.

barmac (approved):

PR #627, opened 2022-03-04
House keeping (dependency bumping)

nikku:
Update dependencies so we're not `CRITICAL` anymore.

Also, really update them.


barmac (approved):
Works great! :rocket:

PR #628, opened 2022-03-09
feat(canvas): add API to hide/show layers

marstamm:
For performance reasons, we only want to have layers of visible object present in the DOM tree. For the reasoning behind it, please see the comment in the platform repo: https://github.com/camunda/camunda-modeler/issues/2803#issuecomment-1059163820

You can now:
 - Hide layers (internal implementation: remove them from the DOM tree)
 - Show layers (internal implementation: add them to the DOM tree again)

By default, every layer is visible. We explicitly have to hide them, e.g. when adding an inactive `rootElement`

### DIFF
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -303,6 +303,25 @@ Canvas.prototype.getLayer = function(name, index) {
 };
 
 /**
+ * For a given index, return the number of layers that have a higher index and
+ * are visible.
+ *
+ * This is used to determine the node a layer should be inserted at.
+ *
+ * @param {Number} index
+ * @returns {Number}
+ */
+Canvas.prototype._getChildIndex = function(index) {
+  return reduce(this._layers, function(childIndex, layer) {
+    if (layer.visible && index >= layer.index) {
+      childIndex++;
+    }
+
+    return childIndex;
+  }, 0);
+};
+
+/**
  * Creates a given layer and returns it.
  *
  * @param {string} name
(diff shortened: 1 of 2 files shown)


lib/core/Canvas.js
```diff
@@ -571,7 +651,7 @@ Canvas.prototype.addRootElement = function(rootElement) {
 
   var layer = this.getLayer(layerName, PLANE_LAYER_INDEX);
 
-  svgAttr(layer, 'display', 'none');
+  this.hideLayer(layerName);
```
nikku, line 654: Idea: Accept `layer` as an argument, too.

test/spec/core/CanvasSpec.js
```diff
@@ -742,31 +742,19 @@ describe('Canvas', function() {
 
     describe('layers', function() {
 
-      it('should require layer name', inject(function(canvas) {
-
-        // then
-        expect(function() {
-          canvas.getLayer();
```
nikku, line 749: What has changed here?

nikku (commented):
Linting needs to be pacified. But looks good other than that.

lib/core/Canvas.js
marstamm: With the API we offer for layers, we never actually return the layer object. We always return the svg group. Using names keeps it consistent with the existing API

marstamm:
> Linting needs to be pacified

True that. Interestingly, it only complains on the pull request trigger but not the push trigger 🤷

marstamm:
probably because I have not rebased to latest master. Will take care of it in the morning

marstamm:
Rebased to develop and fixed linting issues

marstamm:
@nikku , anything still missing before we can merge?

nikku (approved):

PR #630, opened 2022-03-23
fix(context-pad): robust position in line-height != 1 scenarios

nikku:
The current context pad styles are prone to break in environments where line-height is != 1.

I.e. reproducible on `demo.bpmn.io` if `body { line-height: 10; }` is added as a CSS directive.

This PR fixes the behavior; it also ensures that mixed images + icon fonts (appropriately sized) are aligned:

_Chrome_

screenshot 2BZAvz

_Firefox_

screenshot ORLSCZ

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -477,6 +477,7 @@ marker.djs-dragger tspan {
   position: absolute;
   display: none;
   pointer-events: none;
+  line-height: 1;
 }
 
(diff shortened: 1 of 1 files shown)


marstamm (commented):
Nice one! should the target branch be master because it is a fix?

nikku:
Changed the base. Thanks for noticing.

marstamm (approved):

PR #633, opened 2022-03-27
fix(palette): correct separator spacing

nikku:
_This corrects the spacing of the palette separator:_

capture kQmTbX_optimized

---

_Here is an example with a complex palette (mix of icon font (`:before`) + image entries:_

capture z0XWCC_optimized

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -385,7 +385,7 @@ marker.djs-dragger tspan {
 }
 
 .djs-palette .separator {
-  margin: 0 5px;
+  margin: 5px;
   padding-top: 5px;
 
   border: none;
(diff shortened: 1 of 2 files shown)


nikku:
Applications (Camunda Web Modeler, for example) and libraries (i.e. [connectors modeling demo](https://github.com/bpmn-io/bpmn-js-connectors-extension/pull/14/files#diff-c6c9f1543e1d2a53bba79c3ec71198ac1ee229b57e75ce619f76871c4e96564aR11) need to otherwise work around this.

barmac (approved):
image

Looks OK if the extension creators use some transparent padding in their icons.

PR #638, opened 2022-05-17
Add hit box to disable move interaction

barmac:
Use `no-move` type of hit box to disable element movement via hit box.

Related to https://github.com/camunda/camunda-modeler/issues/2859

Test with:

`npx @bpmn-io/sr bpmn-io/bpmn-js#improve-selection-of-a-pool -l bpmn-io/diagram-js#improve-selection-of-a-pool`


nikku (approved):
Great stuff :clap:

Issue #278, opened 2018-09-27
Multi-element context pad


nikku:
Allow a context pad to be displayed for a number of elements.

__Describe the Solution__

* [ ] Context pad API is extended to support multiple elements
* [ ] Selection box + context pad is drawn around multiple elements
 
__Context__

This would allow users to implement batch actions: 

* Bulk property change (https://github.com/camunda/camunda-modeler/issues/892)
* Bulk morph (https://github.com/camunda/camunda-modeler/issues/892)
* Apply color via context pad (single and multi-element)

nikku:
Work in progress exploration of this topic (https://github.com/bpmn-io/diagram-js/tree/multi-element-context-pad).

nikku:
A demo showcasing this end-to-end is [available here](https://github.com/bpmn-io/bpmn-js-color-picker/compare/master...bpmn-io:multi-element-context-pad) (link to respective `multi-element-context-pad` branches in `bpmn-js` and `bpmn-js -> diagram-js`.

__Demo using the feature end-to-end:__

![capture AJc7CN_optimized](https://user-images.githubusercontent.com/58601/140891970-8c58bd60-f05b-4bae-9111-1bb90f1499e4.gif)

PR #587, opened 2021-11-05
Add ability to provide context pads for multiple elements

nikku:
Adds the ability to create a context pad around more than one element:

* [x] dedicated selection box surrounds all selected elements
    * [x] implement
    * [x] test 
* [x] a new context pad provider APIs allows provider to provide pads for more than one element (c.f. https://github.com/bpmn-io/bpmn-js/pull/1525)
    * [x] implement
    * [x] test 
* [x] context pad is now single _and_ multi-element aware. Target now generalized to `djs.model.Base || djs.model.Base[]` accordingly.
* [x] even out popup menu entries so that image only entries are possible (https://github.com/bpmn-io/diagram-js/pull/587/commits/366b5c9190b871031b171d8a5333243f50d96dc3).
* [x] fix selection not updating if selected elements changed (undo/redo)
* [x] test `isOpen` api

---

Capture of the thing in action for "delete" example (with reaction to undo/redo): 

https://user-images.githubusercontent.com/25825387/169019635-83b14574-1a2f-4eaa-b603-645ee357fb50.mov

Delete can be tested by using https://github.com/bpmn-io/bpmn-js/pull/1525 and linking this instance of diagram-js

Closes https://github.com/bpmn-io/diagram-js/issues/278

PR #640, opened 2022-05-18
feat: clearly distinguish select and hover states

nikku:
This proposes a rework of our hover and select outline, as well as interaction (resize, drag, ...) handles:

* No hover state (no blinking as you move the mouse)
* Clear, solid outline for selected elements
* Clear, no opacity, unified size handles to interact with diagram elements (resize, bendpoints, ...)

Role model for this change is Google Slides, but similar approaches can be found in many different apps.

----

capture WUR6jb_optimized

----

Try it out locally:

```
npx @bpmn-io/sr bpmn-io/bpmn-js#develop -l bpmn-io/diagram-js#select-hover-states
```

---

### TODO

* [x] Settle down on variation
* [x] Make this work with multi selection (#278)

----

Related to https://github.com/bpmn-io/bpmn-js/issues/1616

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -97,17 +97,18 @@
 .djs-element.selected .djs-outline {
   visibility: visible;
   shape-rendering: geometricPrecision;
+}
+
+.djs-element.hover .djs-outline {
+  stroke: var(--element-hover-outline-fill-color);
   stroke-dasharray: 3,3;
+  stroke-width: 1px;
 }
 
 .djs-element.selected .djs-outline {
   stroke: var(--element-selected-outline-stroke-color);
   stroke-width: 1px;
-}
-
-.djs-element.hover .djs-outline {
-  stroke: var(--element-hover-outline-fill-color);
-  stroke-width: 1px;
+  stroke-dasharray: none;
 }
 
 .djs-shape.connect-ok .djs-visual > :nth-child(1) {
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -814,3 +814,9 @@ marker.djs-dragger tspan {
 .djs-label-hidden .djs-label {
   display: none !important;
 }
+
+.djs-element .djs-hit-stroke,
+.djs-element .djs-hit-click-stroke,
+.djs-element .djs-hit-all {
+  cursor: move;
+}
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -15,6 +15,7 @@
   --color-blue-205-100-45: hsl(205, 100%, 45%);
   --color-blue-205-100-45-opacity-30: hsla(205, 100%, 45%, 30%);
   --color-blue-205-100-50: hsl(205, 100%, 50%);
+  --color-blue-205-100-50-opacity-15: hsla(205, 100%, 50%, 15%);
   --color-blue-205-100-70: hsl(205, 100%, 75%);
   --color-blue-205-100-95: hsl(205, 100%, 95%);
 
@@ -27,7 +28,6 @@
 
   --color-white: hsl(0, 0%, 100%);
   --color-black: hsl(0, 0%, 0%);
-  --color-black-opacity-05: hsla(0, 0%, 0%, 5%);
   --color-black-opacity-10: hsla(0, 0%, 0%, 10%);
 
   --canvas-fill-color: var(--color-white);
@@ -43,8 +43,8 @@
   --element-selected-outline-stroke-color: var(--color-blue-205-100-50);
   --element-selected-outline-secondary-stroke-color: var(--color-blue-205-100-70);
 
-  --lasso-fill-color: var(--color-black-opacity-05);
-  --lasso-stroke-color: var(--color-black);
+  --lasso-fill-color: var(--color-blue-205-100-50-opacity-15);
+  --lasso-stroke-color: var(--element-selected-outline-stroke-color);
 
   --palette-entry-color: var(--color-grey-225-10-15);
   --palette-entry-hover-color: var(--color-blue-205-100-45);
@@ -171,10 +171,8 @@ svg.new-parent {
 */
 .djs-lasso-overlay {
   fill: var(--lasso-fill-color);
-
-  stroke-dasharray: 5 1 3 1;
   stroke: var(--lasso-stroke-color);
-
+  stroke-width: 2px;
   shape-rendering: geometricPrecision;
   pointer-events: none;
 }
@@ -200,7 +198,7 @@ svg.new-parent {
   fill: var(--resizer-fill-color);
   stroke-width: 1px;
   stroke: var(--resizer-stroke-color);
-  shape-rendering: geometricprecision;
+  shape-rendering: geometricPrecision;
 }
 
 .djs-resizer:hover .djs-resizer-visual {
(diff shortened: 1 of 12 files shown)


philippfromme:
In my opinion we should not only increase the line width but also the size of the handles (resize, segment move), which are smaller then bendpoint move handles. I proposed this years ago.

brave_oFzsDMFUMP

andreasgeier:
I'd go with 2px. Consider, that single select is the default case and it should help to better see/find the selected element. Particularly in case of clicking on an error message references.

Please avoid a hover state on selected elements. It's not common and I believe it is confusing.

nikku:
@philippfromme You have your adjustments on a branch?

philippfromme:
@nikku https://github.com/bpmn-io/diagram-js/tree/philipp-styles 😉

christian-konrad:
2px looks better. 1px makes sense for high-precision drawing tools (adobe, affinity, sketch...), but for a rapid modeling tool, quick distinguishability and clickability is more important.

assets/diagram-js.css
```diff
@@ -97,17 +97,18 @@
...
+  stroke: var(--element-hover-outline-fill-color);
+  stroke-dasharray: 3,3;
   stroke-width: 1px;
 }
 
-.djs-element.hover .djs-outline {
-  stroke: var(--element-hover-outline-fill-color);
+.djs-element.selected .djs-outline {
+  stroke: var(--element-selected-outline-stroke-color);
   stroke-width: 1px;
```
christian-konrad, line 110: 2px as discussed

christian-konrad (changes requested):

nikku:
I'll provide another version that incorporates @philippfromme styles. If we go for https://github.com/bpmn-io/diagram-js/pull/640#issuecomment-1130003235, more general clarity on our UI controls helps.

nikku:
> 
2px looks better. 1px makes sense for high-precision drawing tools (adobe, affinity, sketch...), but for a rapid modeling tool, quick distinguishability and clickability is more important.

Visual sizing has no impact for `clickability`. This is really about what we want to be, a happy looking sketching tool, or a slight less in your face app. Miro for example goes for 1px, no problem.

nikku:
Accounting for @philippfromme feedback but also taking into consideration other drawing apps I've reworked this PR:

* Outline is solid, if it is shown
* Hover outline is gone (no-one has it)
* Handles are not transparent, and sized appropriately
* No two pixel border, as this looks toyish (and other tools don't have it, too)

---

I've considered the following tools for comparison:

* Google Slides
* Excalidraw
* Miro

currandwyer:
I think this is looking great! Cool that we're working on it. One question for you @nikku - why don't we ease the corners of the selection boxes? I think this looks much more refined, given that almost everything being selected also has rounded corners. All we would need to do is decide on a different behavior for pools, which seems doable to me.

philippfromme:
> why don't we ease the corners of the selection boxes?

@currandwyer  Do you mean border radius?

nikku:
@philippfromme That is a minimal one, right? I'll give it a shot and report back how that looks like.

currandwyer:
Yeah @philippfromme I think even just having 1-2 px corner radius is nice - it's subtle, but it softens the look a bit, and shows that you thought about it. By the way, in Operate we already have a relatively large corner radius:
Screen Shot 2022-04-20 at 19 21 02

nikku:
@currandwyer Thanks for the operate screen. Does the user interact with Participants there? How does operate handle selection on DMN DRD diagrams (with decisions and text annotations)?

image

nikku:
And as we're on it, here is a "bold selection outline" (3px radius) variation: 

capture mMaabe_optimized

nikku:
One thing we have to consider is the newly introduced multi-selection box (https://github.com/bpmn-io/diagram-js/pull/587).

nikku:
Marked as draft as we clearly still discover what we want to do :wink:

nikku:
Miro does something else interesting which we'll need to account for now, too (https://github.com/bpmn-io/diagram-js/pull/587). They color elements differently: Distinguish primary from secondary selection:

capture a51Lxn_optimized

philippfromme:
> What do you think?

Minimal border radius (2-3px) would be my choice. If it's too big it looks weird and distracting.

philippfromme:
> And as we're on it, here is a "bold selection outline"

I'm a fan of 2px stroke width. It looks so satisfying. 😋

philippfromme:
> @currandwyer Thanks for the operate screen. Does the user interact with Participants there? How does operate handle selection on DMN DRD diagrams (with decisions and text annotations)?

This is why I'd say minimal border radius is fine. Due to the padding it would never overlap the element.

currandwyer:
I think that 3px radius is great! @nikku I wasn't being clear about Operate - I am wondering why it's done differently. To me, it seems like it should be aligned.

nikku:
Proposal for solution, incorporating modern hover vs. select interaction; clearly distinguishable (+ rounded nice to the eye) selection box + multi-select aware:

capture WUR6jb_optimized

philippfromme:
> Proposal for solution, incorporating modern hover vs. select interaction; clearly distinguishable (+ rounded nice to the eye) selection box + multi-select aware:

I'm a fan. 💯

philippfromme:
I was wondering, shouldn't the outline always be on top? 🤔 

image

christian-konrad:
@nikku  is it possible to have the outline thickness be fixed, so not to scale with the canvas when you zoom?

christian-konrad:
@nikku interesting thought to get rid of the hover outline at all, this solves many problems with multi-select.
But, would it be possible then to turn the cursor into a drag cursor (like in Google Slides / Excalidraw)
(Just a thought, I am not even sure if I like it)

nikku:
> @nikku is it possible (and to what cost) to have the outline thickness be fixed, so not to scale with the canvas when you zoom?

Not easily possible.

> But, would it be possible then to turn the cursor into a drag cursor. 

When exactly? We could do it. But I don't miss it (and no-one ever did).
It would also cause conflicts, i.e. in places where you could both start dragging, but also direct editing via double click. TLDR: Let's not optimize it in this PR :wink:

nikku:
> I was wondering, shouldn't the outline always be on top? thinking Like, not covered by the boundary event.

Let us follow up with this in a new issue @philippfromme.

I've checked other tools and that is exactly how they do it. Our multi-select outline also works like this: 

image

philippfromme:
> Let us follow up with this in a new issue @philippfromme.

👍🏻

christian-konrad:
Just found this bug while testing, which could be caused by this https://github.com/bpmn-io/bpmn-js/issues/1657

christian-konrad:
I am not 100% happy with having no hover state anymore, but as this solution solves current problems, please move forward with it.
Reconsider hovering if a designer comes up with it and good reasoning.
(There isn't even a real best practice there, there are vector, drawing, model etc tools with and without hover states)

christian-konrad (approved):

nikku:
Potential follow up once we got this merged (https://github.com/bpmn-io/diagram-js/pull/643), incorporating @christian-konrad earlier comment (https://github.com/bpmn-io/diagram-js/pull/640#issuecomment-1131903966).

nikku:
@philippfromme As discussed, please leave this.

nikku:
Hahaha. leave you review :).

philippfromme (approved):
This is such a neat package of improvements! 🏅 I love it!

PR #644, opened 2022-05-20
Select connection

nikku:
This fixes what bugged me for a long time: The fact that we do not select newly created connections, but rather the target of the connect operation. 

capture kjTAff_optimized

As you see from the commits we were not even ready for this "fixed interaction".

----

Try it out via 

```
npx @bpmn-io/sr bpmn-io/bpmn-js#develop -l bpmn-io/diagram-js#select-connection
```


MaxTru (approved):
* Works (manually integration tested)
* Commits help to understand the change, smooth
* Lets have this!

PR #646, opened 2022-05-23
feat(snapping): snap bendpoints on connection path

marstamm:
related to https://github.com/bpmn-io/bpmn-js/pull/1661

### DIFF
--- a/lib/features/bendpoints/BendpointSnapping.js
+++ b/lib/features/bendpoints/BendpointSnapping.js
@@ -6,6 +6,8 @@ import {
 
 import { setSnapped } from '../snapping/SnapUtil';
 
+import { getClosestPointOnConnection } from './BendpointUtil';
+
 var abs = Math.abs,
     round = Math.round;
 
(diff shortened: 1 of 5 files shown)


nikku (changes requested):
I understand what we're doing here. I'm wondering though if "cropping" is the right thing to do for connections.

Connecting to shapes there is a _docking point_ and a actual target point. The connection is _cropped_ visually, but we still show the target point (as the bendpoint).


Connecting to connections I think there should be simply one point, the target point, and it should snap to the connection (no cropping!). Cf. the following screencapture that shows this:

capture FQYBhq_optimized

marstamm:
> I'm wondering though if "cropping" is the right thing to do for connections

Is this just about the terminology used in this utility (e.g., we should rename it to `getDockedWaypoints`)? As I understand it, the current behavior is already how you describe it, extending or shortening the line to match:

Recording 2022-05-23 at 13 57 44

marstamm:
Now using actual snapping! Try it out with

`npx @bpmn-io/sr bpmn-io/bpmn-js#develop -l bpmn-io/diagram-js#1660-snap-to-flow`

I'll keep https://github.com/bpmn-io/bpmn-js/pull/1661 open to refactor the GeometricUtil when we released `diagram-js`, but all logic is in diagram-js.

Recording 2022-05-23 at 16 44 07

PR #647, opened 2022-05-24
feat(css): remove higlighting for connections

marstamm:
Instead of making the Connection nearly invisible, use cursor style to show allowed drop targets. This is consistent with shapes, as the outline is not changed when highlighted.

We can think about another highlight option in a follow-up.

Recording 2022-05-24 at 12 38 36


nikku (approved):
I like this as a first step to _fix things_.

PR #648, opened 2022-05-24
fix(modeling): remove incoming/outgoing on `connection.delete`

marstamm:
Recording 2022-05-24 at 14 12 21

related to https://github.com/bpmn-io/bpmn-js/issues/1663

### DIFF
--- a/test/spec/features/modeling/DeleteConnectionSpec.js
+++ b/test/spec/features/modeling/DeleteConnectionSpec.js
@@ -15,7 +15,8 @@ describe('features/modeling - remove connection', function() {
   }));
 
 
-  var rootShape, parentShape, childShape, childShape2, connection;
+  var rootShape, parentShape, childShape, childShape2,
+      connection, outgoingConnection, incommingConnection;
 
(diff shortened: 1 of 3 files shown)


test/spec/features/modeling/DeleteConnectionSpec.js
```diff
@@ -67,6 +87,19 @@ describe('features/modeling - remove connection', function() {
   }));
 
 
+  it('should remove Outgoing', inject(function(modeling, elementRegistry) {
```
nikku, line 90: Should we add a test case to verify batch deletion works ("no double deletion happens")?

nikku (changes requested):
Let's add one more [test case](https://github.com/bpmn-io/diagram-js/pull/648/files#r880470077)

test/spec/features/modeling/DeleteConnectionSpec.js
nikku: Should be in `DeleteElementsSpec` though.

nikku:
:clap

PR #651, opened 2022-06-10
fix: limit size of context pad entry image

barmac:
This is to make sure that I can provide SVGs without worrying about the icon size.

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -501,11 +501,6 @@ marker.djs-dragger tspan {
   background: var(--context-pad-entry-hover-background-color);
 }
 
-.djs-context-pad .entry img {
-  max-width: 100%;
-  max-height: 100%;
-}
-
 .djs-context-pad.open {
   display: block;
 }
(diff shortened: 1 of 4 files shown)


nikku (changes requested):
Could you add an example to the existing context pad specs to verify this (visually)?

barmac:
I tried to but then it occurred that none of the images in the current example are displayed correctly :D Let me push those changes.

barmac:
Hmm it seems to work. I will add a proper test.

barmac:
OK I've added a test case. This will be required for the alignment icons.

barmac:
Learning: image loads asynchronously so we need to wait before we compare size.

nikku (commented):
Works for me. 

An alternative is to enforce the correct size directly, did you consider it?

barmac:
> Works for me.
> 
> An alternative is to enforce the correct size directly, did you consider it?

Via `style`? I have not. In what way could it improve the solution?

nikku:
> In what way could it improve the solution?

It would enforce a consistent sizing, not just max sizing. Generally I'd go an extra step to ensure users do not shoot themselves in the foot.

nikku (approved):

PR #653, opened 2022-06-27
Group popup menu items

barmac:
Required for https://github.com/bpmn-io/bpmn-js/pull/1682

### DIFF
--- a/lib/features/popup-menu/PopupMenu.js
+++ b/lib/features/popup-menu/PopupMenu.js
@@ -509,7 +509,9 @@ PopupMenu.prototype._createEntries = function(entries, className) {
         groupContainer = domQuery('[data-group=' + grouping + ']', entriesContainer);
 
     if (!groupContainer) {
-      groupContainer = domify('<div class="group" data-group="' + grouping + '"></div>');
+      groupContainer = domify('<div class="group"></div>');
+      domAttr(groupContainer, 'data-group', grouping);
+
       entriesContainer.appendChild(groupContainer);
     }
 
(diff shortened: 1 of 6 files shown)


lib/features/popup-menu/PopupMenu.js
```diff
@@ -501,8 +504,16 @@ PopupMenu.prototype._createEntries = function(entries, className) {
...
 
   forEach(entries, function(entry, id) {
-    var entryContainer = self._createEntry(entry, id);
-    entriesContainer.appendChild(entryContainer);
+    var entryContainer = self._createEntry(entry, id),
+        grouping = entry.group || 'default',
+        groupContainer = domQuery('[data-group=' + grouping + ']', entriesContainer);
+
+    if (!groupContainer) {
+      groupContainer = domify('<div class="group" data-group="' + grouping + '"></div>');
```
nikku, line 512: We should escape `grouping` to not allow XSS.

nikku (changes requested):

lib/features/popup-menu/PopupMenu.js
barmac: Valid point. We will need to fix it in [the context pad](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/context-pad/ContextPad.js#L286) as well.

barmac: This is now fixed, but there are also a couple of other places where we allow DOM injection. I will fix them.

nikku: Woops woops :).

PR #654, opened 2022-06-28
fix(popup-menu): display even if no cursor is passed to `open`

barmac:
Previously, the popup menu would be stuck with `visibility: hidden`.


nikku (approved):
Good catch. Visibility shall not depend on actual cursor position.

PR #656, opened 2022-07-01
Use rules for alignment and distribution

barmac:


marstamm (approved):
Love the use of rules and getting rid of the magic "filter big elements" code

PR #657, opened 2022-07-06
fix(util): properly assign helper-svg styles

nikku:
Related to https://github.com/bpmn-io/bpmn-js/issues/1692

### DIFF
--- a/lib/util/Text.js
+++ b/lib/util/Text.js
@@ -12,6 +12,10 @@ import {
   remove as svgRemove
 } from 'tiny-svg';
 
+import {
+  assignStyle
+} from 'min-dom';
+
 var DEFAULT_BOX_PADDING = 0;
 
 var DEFAULT_LABEL_SIZE = {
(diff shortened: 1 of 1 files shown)


barmac:
I can't reproduce the original issue on Mac. @philippfromme @Skaiir one of you could probably help out here.

nikku:
Interesting that this is not an issue on MacOS? I guess you're also using latest Chrome to test?


You should however be able to reproduce the fact that the original code (fixed) does not do what it is supposed to do.

Assigning `visibility: none` to a DOM element does nothing. This works for SVG only, sometimes.

barmac:
> Interesting that this is not an issue on MacOS? I guess you're also using latest Chrome to test?

I checked once more on the demo site and I could reproduce the problem to some extent. It required that I scrolled while hovering the text on the bottom or the bpmn.io logo.

Scrollbars are not normally visible on Mac, but they appear when I initiate scrolling. This was possible before your changes and impossible after they were applied.

barmac (approved):
This fixes the issue. Each of the style properties is required for the fix to work.

barmac:
Hmm this could have been merged to `master`.

nikku:
:see_no_evil:

PR #660, opened 2022-08-15
feat(editor-actions): only copy if selected elements

nikku:
Also, properly indicate copy result ("did I copy?").

----

This ensures we can use state of the art copy behavior consistently across our lib (i.e. also in cross browser copy + paste: https://github.com/nikku/bpmn-js-copy-paste-example)


philippfromme (approved):
✅

Issue #661, opened 2022-08-15
Make `keyboard` binding implicit / remove ability to bind to any element


nikku:
__Is your feature request related to a problem? Please describe.__

For historical reasons our users had to bind `keyboard` to any element on the page, may it be document, or the diagram container. With https://github.com/bpmn-io/diagram-js/pull/662 we make the diagram canvas browser selectable; as such we can safely figure out if the canvas has focus and keyboard bindings shall be accounted for.

Because of that we want to remove the bind anywhere `keyboard#bindTo` / `keyboard#bind(element)` APIs.

__Describe the solution you'd like__

* Keyboard can be bound and unbound, _however_ it will always be bound to the canvas
* Binding keyboard to a particular element is removed
* Configuring element to be bound to via `keyboard.bindTo` is removed
* Proper error messages indicate old use
* API to focus the diagram (`canvas.focus()`, and potentially `diagram.focus()` as a proxy) exist
* To be considered: The canvas fires `focus` and `blur` events that I can hook into to register global keyboard shortcuts
* Once the canvas is focused, then we spend reasonable amount of effort to keep it focused / restore focus once on canvas edit operations happened (https://github.com/bpmn-io/diagram-js/pull/662#issuecomment-2314722118)

__Describe alternatives you've considered__

Keep stuff and magic as is.

__Additional context__

This is a major improvement UX wise.

marstamm:
@nikku , assigned to you for now to clarify what is needed to enable HTO team

nikku:
@vsgoulart I'll not be able to follow up on this in the next days, but will revisit end of March.

barmac:
Assigning to @jarekdanielak to assess in a timebox manner.

nikku:
Explanatory scetch, based of discussion with @jarekdanielak:

![image](https://github.com/user-attachments/assets/98180f0a-1c18-474f-a6bc-148fdc4b8b16)

nikku:
Existing ways to indicate breaking changes (APIs no longer working or changed) - https://github.com/search?q=repo%3Abpmn-io%2Fdiagram-js%20console.warn&type=code.

nikku:
Other editors, e.g. [code mirror](https://codemirror.net/) are great examples for friendly, and accessible integration into the browser.

PR #662, opened 2022-08-15
Make canvas selectable / keyboard binding implicit

nikku:
Keyboard is not implicitly bound and `Canvas` is a focusable element.

This builds foundations to [support cross-browser/application copy and paste](https://github.com/bpmn-io/internal-docs/issues/614) and more general keyboard related UX enhancements.

This change introduces a predictable editing experience, simplifies the keyboard and ensures worry free interoperability with multiple editors and inputs on the page.

### Keyboard binding (https://github.com/bpmn-io/diagram-js/pull/662/commits/b0f9a75fc2cb38a1a973763512821c92550837db)

Keyboard is now implicitly bound to the canvas element (svg). It provides behavior closer to native browser experience.

Previous calls to `keyboard.bind(someElement)` or configuration via `keyboard.bindTo` log a descriptive descriptive error:

image

### Canvas focusable (https://github.com/bpmn-io/diagram-js/pull/662/commits/c75285be7eb40da0de5a193bfe95f03865c52fd0)

The `Canvas` is now a focusable component, that is recognized accordingly by the browser, with all benefits for UX and interaction. New APIs are provided to allow restoring focus to the canvas.

Components that pull focus from the `Canvas` during modeling must ensure to restore the focus (if intended), via `Canvas#restoreFocus`.

---

Closes https://github.com/bpmn-io/diagram-js/issues/661

### DIFF
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -10,7 +10,8 @@ import {
 } from 'min-dash';
 
 import {
-  assignStyle
+  assignStyle,
+  attr as domAttr
 } from 'min-dom';
 
 import {
@@ -189,6 +190,11 @@ export default function Canvas(config, eventBus, graphicsFactory, elementRegistr
    */
   this._rootElement = null;
 
+  /**
+   * @type {boolean}
+   */
+  this._focused = false;
+
   this._init(config || {});
 }
 
@@ -215,14 +221,33 @@ Canvas.$inject = [
  * @param {CanvasConfig} config
  */
 Canvas.prototype._init = function(config) {
-
   const eventBus = this._eventBus;
 
   // html container
   const container = this._container = createContainer(config);
 
   const svg = this._svg = svgCreate('svg');
-  svgAttr(svg, { width: '100%', height: '100%' });
+
+  svgAttr(svg, {
+    width: '100%',
+    height: '100%'
+  });
+
+  domAttr(svg, 'tabindex', 0);
+
+  eventBus.on('element.hover', () => {
+    this.restoreFocus();
+  });
+
+  svg.addEventListener('focusin', () => {
+    this._focused = true;
+    eventBus.fire('canvas.focus.changed', { focused: true });
+  });
+
+  svg.addEventListener('focusout', () => {
+    this._focused = false;
+    eventBus.fire('canvas.focus.changed', { focused: false });
+  });
 
   svgAppend(container, svg);
 
@@ -312,6 +337,31 @@ Canvas.prototype._clear = function() {
 
   // force recomputation of view box
   delete this._cachedViewbox;
+};
+
+/**
+* Sets focus on the canvas SVG element.
+*/
+Canvas.prototype.focus = function() {
+  this._svg.focus({ preventScroll: true });
+};
+
+/**
+* Sets focus on the canvas SVG element if `document.body` is currently focused.
+*/
+Canvas.prototype.restoreFocus = function() {
+  if (document.activeElement === document.body) {
+    this.focus();
+  }
+};
+
+/**
+* Returns true if the canvas is focused.
+*
+* @return {boolean}
+*/
+Canvas.prototype.isFocused = function() {
+  return this._focused;
 };
 
 /**
--- a/lib/features/keyboard/Keyboard.js
+++ b/lib/features/keyboard/Keyboard.js
@@ -3,9 +3,7 @@ import {
 } from 'min-dash';
 
 import {
-  closest as domClosest,
-  event as domEvent,
-  matches as domMatches
+  event as domEvent
 } from 'min-dom';
 
 import {
@@ -24,9 +22,10 @@ import {
 var KEYDOWN_EVENT = 'keyboard.keydown',
     KEYUP_EVENT = 'keyboard.keyup';
 
-var HANDLE_MODIFIER_ATTRIBUTE = 'input-handle-modified-keys';
-
 var DEFAULT_PRIORITY = 1000;
+
+var compatMessage = 'Keyboard binding is now implicit; explicit binding to an element got removed. For more information, see https://github.com/bpmn-io/diagram-js/issues/661';
+
 
 /**
  * A keyboard abstraction that may be activated and
@@ -46,8 +45,8 @@ var DEFAULT_PRIORITY = 1000;
  *
  * All events contain one field which is node.
  *
- * A default binding for the keyboard may be specified via the
- * `keyboard.bindTo` configuration option.
+ * Specify the initial keyboard binding state via the
+ * `keyboard.bind=true|false` configuration option.
  *
  * @param {Object} config
  * @param {EventTarget} [config.bindTo]
@@ -56,7 +55,8 @@ var DEFAULT_PRIORITY = 1000;
 export default function Keyboard(config, eventBus) {
   var self = this;
 
-  this._config = config || {};
+  this._config = config = config || {};
+
   this._eventBus = eventBus;
 
   this._keydownHandler = this._keydownHandler.bind(this);
@@ -69,19 +69,22 @@ export default function Keyboard(config, eventBus) {
     self.unbind();
   });
 
-  eventBus.on('diagram.init', function() {
+  if (config.bindTo) {
+    console.error('unsupported configuration <keyboard.bindTo>', new Error(compatMessage));
+  }
+
+  var bind = config && config.bind !== false;
+
+  eventBus.on('canvas.init', function(event) {
+    self._target = event.svg;
+
+    if (bind) {
+      self.bind();
+    }
+
     self._fire('init');
   });
 
-  eventBus.on('attach', function() {
-    if (config && config.bindTo) {
-      self.bind(config.bindTo);
-    }
-  });
-
-  eventBus.on('detach', function() {
-    self.unbind();
-  });
 }
 
 Keyboard.$inject = [
@@ -116,34 +119,7 @@ Keyboard.prototype._keyHandler = function(event, type) {
 };
 
 Keyboard.prototype._isEventIgnored = function(event) {
-  if (event.defaultPrevented) {
-    return true;
-  }
-
-  return (
-    isInput(event.target) || (
-      isButton(event.target) && isKey([ ' ', 'Enter' ], event)
-    )
-  ) && this._isModifiedKeyIgnored(event);
-};
-
-Keyboard.prototype._isModifiedKeyIgnored = function(event) {
-  if (!isCmd(event)) {
-    return true;
-  }
-
-  var allowedModifiers = this._getAllowedModifiers(event.target);
-  return allowedModifiers.indexOf(event.key) === -1;
-};
-
-Keyboard.prototype._getAllowedModifiers = function(element) {
-  var modifierContainer = domClosest(element, '[' + HANDLE_MODIFIER_ATTRIBUTE + ']', true);
-
-  if (!modifierContainer || (this._node && !this._node.contains(modifierContainer))) {
-    return [];
-  }
-
-  return modifierContainer.getAttribute(HANDLE_MODIFIER_ATTRIBUTE).split(',');
+  return false;
 };
 
 /**
@@ -153,10 +129,14 @@ Keyboard.prototype._getAllowedModifiers = function(element) {
  */
 Keyboard.prototype.bind = function(node) {
 
+  if (node) {
+    console.error('unsupported argument <node>', new Error(compatMessage));
+  }
+
   // make sure that the keyboard is only bound once to the DOM
   this.unbind();
 
-  this._node = node;
+  node = this._node = this._target;
 
   // bind key events
   domEvent.bind(node, 'keydown', this._keydownHandler);
@@ -226,15 +206,3 @@ Keyboard.prototype.hasModifier = hasModifier;
 Keyboard.prototype.isCmd = isCmd;
 Keyboard.prototype.isShift = isShift;
 Keyboard.prototype.isKey = isKey;
-
-
-
-// helpers ///////
-
-function isInput(target) {
-  return target && (domMatches(target, 'input, textarea') || target.contentEditable === 'true');
-}
-
-function isButton(target) {
-  return target && domMatches(target, 'button, input[type=submit], input[type=button], a[href], [aria-role=button]');
-}
--- a/lib/features/popup-menu/PopupMenu.js
+++ b/lib/features/popup-menu/PopupMenu.js
@@ -262,6 +262,8 @@ PopupMenu.prototype.close = function() {
 
   this.reset();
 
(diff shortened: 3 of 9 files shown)


nikku:
This can be tried out via: 

```
npx @bpmn-io/sr bpmn-io/bpmn-js#keyboard-bind -l bpmn-io/diagram-js#keyboard-bind -c "npm run start"
```

----

What I did is to remove any input quirks but rely on canvas to be actually browser selected. As a side-effect though that means that we consistently need to select the browser. I do not see a simple way to accomplish this.

nikku:
The following shows how we may mitigate this: Whenever the canvas is mouse hovered we re-focus the canvas.

For debugging purposes I added a solid blue focus outline (solid blue => receives keyboard events + can copy + paste).

capture 3jv7hP_optimized

barmac:
On MacOS, the focus state is persisted once I hover the canvas even if switch tab afterwards and come back. Is this is supposed to work like this? I am not opinionated but it would be logical to me that focus fades away when I move mouse outside without clicking.

barmac:
Cf. recording

https://user-images.githubusercontent.com/28307541/184912457-4cb8059b-2bf0-4027-912d-52aba7155da7.mov

nikku:
> https://github.com/bpmn-io/diagram-js/pull/662#issuecomment-1216757684

I guess so. Keyboard inputs also don't magically unfocus once you leave with the mouse. The existing implementation works in the exact same way, or at least should work in the same way.

barmac:
Hmm I don't like the infinite scrolling of the canvas which triggers once you paste outside of the canvas, but this solution is still an improvement compared to what we have right now. Thus, 👍 from me.

nikku:
Fortunately infinite scrolling is an existing feature (cf. demo.bpmn.io):

capture e4fk9B_optimized

marstamm:
When we implicitly bind to the canvas, does it imply that extensions that rely on the command stack need to catch and trigger undo/redo manually? I'm thinking about the Properties panel here, where the most common case was to bind the keyboard to a common container.

I tested it on Edge and it works as expected, no new issues found from my side

nikku:
> When we implicitly bind to the canvas, does it imply that extensions that rely on the command stack need to catch and trigger undo/redo manually?

Exactly. The related hacks are all gone from the code base. If you want to trigger undo/redo via keyboard shortcuts you _have to_ implement that yourself (i.e. attach a respective listener on the properties panel).

nikku:
@philippfromme Given that there is a little bit of focus/unfocus going on we could decide to ditch the focus outline. What do you think?

barmac:
While I was able to break the focus-on-hover on Safari, it's due to some `element.hover` bug – which is IMO an edge case we shouldn't care about.

https://user-images.githubusercontent.com/28307541/185079281-690ceb1e-c49f-43b3-9e37-06319e212a02.mov

---

Regarding the cross-browser copy and paste, on MacOS I am able to copy and paste across single browser tabs or even windows, but I cannot do that between different browsers (Safari -> Chrome, Chrome -> Safari). The clipboardData appeared to be empty when I tried to debug it.

nikku:
Found the following issues integration testing this against different libraries in our eco-system:

* dmn-js-drd: no issues (works out of the box)
* dmn-js-decision-table / literal expression editor: has it's own keyboard implementation :arrow_right: needs porting to similar bind behavior, i.e. either make the table selectable + bind to it. 
* form-js: re-uses diagram-js keyboard but has no canvas :arrow_right: need to improve the bind / unbind behavior proposed in this PR to get the actual diagram container. Two ideas: 
     * expose diagram with container in form-js, too
     * pass the container on diagram.init for components to catch it and do stuff with it (i.e . register/unregister key bindings)

pinussilvestrus:
> form-js: re-uses diagram-js keyboard but has no canvas ➡️ need to improve the bind / unbind behavior proposed in this PR to get the actual diagram container. Two ideas:
> * expose diagram with container in form-js, too
> * pass the container on diagram.init for components to catch it and do stuff with it (i.e . register/unregister key bindings)

Given we follow one of these approaches, would we also have to follow up with explicit keyboard bindings for the Forms Properties Panel? 

I'm asking because we plan to make it independent from the editor (e.g. render it in an own container, as we do with the other properties panels, cf. https://github.com/bpmn-io/form-js/pull/291).

nikku:
> [...] would we also have to follow up with explicit keyboard bindings for the Forms Properties Panel?

Absolutely. Taking it one step at a time here though. Blueprint for properties panel binding can be found here: https://github.com/bpmn-io/bpmn-js-properties-panel/pull/739.

nikku:
This is now available as a pre-release via `diagram-js@9.0.0-alpha.1` and `bpmn-js@10.0.0-alpha.1`.

nikku:
Found one issue integration testing this:

capture VmbO2q_optimized

From what it looks like we need to prevent the built in scroll on focus behavior (https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus).

nikku:
Addressed https://github.com/bpmn-io/diagram-js/pull/662#issuecomment-1218241894 via https://github.com/bpmn-io/diagram-js/pull/662/commits/bdea9c9cceedcbec4f0423626b3f2d12edd904de

nikku:
[The form-js integration](https://github.com/bpmn-io/form-js/pull/311), specifically the playground shows best what browser native focus handling + implicit keyboard binding can accomplish:


capture CPUDfT_optimized

pinussilvestrus:
> [The form-js integration](https://github.com/bpmn-io/form-js/pull/311), specifically the playground shows best what browser native focus handling + implicit keyboard binding can accomplish:
> 

Nice, so it fixes https://github.com/bpmn-io/form-js/issues/277 👍

vsgoulart:
I rebased this branch with the current `main`, I saved the previous state in [this other branch](https://github.com/bpmn-io/diagram-js/tree/original-keyboard-bind) since we had some conflicts while rebasing it

vsgoulart:
@nikku do you have a (non-binding) ETA for this PR? :sweat_smile: 

Just so I can update the team

nikku:
@vsgoulart I'm going to assess it this week.

jarekdanielak:
The solution with focus on hover works well. I will prepare alpha release and work on integrating it down the stream.

Known issues:
* When typing in label/annotation, if you finish with Return Key while the cursor is not in the canvas, it won't focus.
* When interacting with bottom panel (problems, output, plugins etc.), canvas won't re-focus on hover as the focus is not on `document.body`.

lib/core/Canvas.js
```diff
@@ -229,8 +228,14 @@ Canvas.prototype._init = function(config) {
     height: '100%'
   });
 
+  assignStyle(svg, {
```
nikku, line 231: [q]: Why do we do this here, rather than via CSS?

lib/core/Canvas.js
```diff
@@ -229,8 +228,14 @@ Canvas.prototype._init = function(config) {
...
+    outline: 'none'
+  });
+
   domAttr(svg, 'tabindex', 0);
 
+  const focus = this.focus.bind(this);
```
nikku, line 237: Why do we do this? Usually we don't do `bind` shinanigans, but ask users to properly execute the actual method with the receiver, i.e. `canvas.focus()`.

lib/features/popup-menu/PopupMenu.js
```diff
@@ -262,6 +262,8 @@ PopupMenu.prototype.close = function() {
 
   this.reset();
 
+  this._canvas.focus();
```
nikku, line 265: Proposal: Consider if an element other than the document was focused, and if not, restore canvas focus.

Rationale: A popup menu action _may_, i.e. open a modal, and we don't want to take the focus if the modal was focused.

TODO: Also write a test case for such a scenario.

lib/core/Canvas.js
```diff
@@ -327,14 +332,10 @@ Canvas.prototype._clear = function() {
   delete this._cachedViewbox;
 };
 
-Canvas.prototype._focus = function() {
+Canvas.prototype.focus = function() {
```
nikku, line 335: Great to have a public API for this. 

I very naturally want to control the focus state, same as in CodeMirror or any other editor.

lib/core/Canvas.js
jarekdanielak: [Camunda Modeler style](https://github.com/camunda/camunda-modeler/blob/5574871edd34b5ba83ff97bc9d30206456db64e3/client/src/styles/_base.less#L17) was overruling any attempt to set it in diagram-js other than inline like that.

lib/core/Canvas.js
jarekdanielak: Leftover. Removed.

lib/core/Canvas.js
jarekdanielak: Added some JSDoc and generated types.

lib/features/popup-menu/PopupMenu.js
jarekdanielak: Done 👍

Do you have an idea how to mock menu closing into some modal-like container?

nikku: Create a DOM element outside of the Canvas. Focus it as part of the popup menu action. Observe the result.

lib/core/Canvas.js
```diff
@@ -190,6 +190,11 @@ export default function Canvas(config, eventBus, graphicsFactory, elementRegistr
    */
   this._rootElement = null;
 
+  /**
+   * @type {boolean}
+   */
+  this._focused = false;
```
nikku, line 196: Let's also expose this as an API.

nikku: Useful in the following scenario:

```
eventBus.on([ 'canvas.focusin', 'canvas.focusout' ], () => {
  if (canvas.isFocussed()) {
    // bind keyboard listeners
  } else {
    // unbind keyboard listeners
  }
});
```

Alternatively we can better support this via a more general event `canvas.focus.changed`:

```
eventBus.on([ 'canvas.focus.changed' ], (event) => {
  if (event.focussed) {
    // bind keyboard listeners
  } else {
    // unbind keyboard listeners
  }
});
```

nikku: The later one will simplify things, and better align with `canvas.viewbox.changed` that we already emit.

lib/core/Canvas.js
nikku: I still don't understand. We should offer stable defaults, and allow tools, i.e. Camunda Modeler to do exactly that customization.

If Camunda Modeler chooses to overrule things, then may it be, if the modeler chooses to exclude the canvas from the outline, then a simple `:not(.canvas):focus { ... }` selector is all it needs.

lib/core/Canvas.js
```diff
@@ -347,9 +345,23 @@ Canvas.prototype._clear = function() {
 
 /**
 * Sets focus on the canvas SVG element.
+*
+* @param {boolean} [onlyFromBody=true] If true, will only focus canvas if
+*     current active element is `document.body`. True by default.
+*/
+Canvas.prototype.focus = function(onlyFromBody = true) {
```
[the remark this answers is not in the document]
jarekdanielak: I have split the focus logic into `focus` and `restoreFocus` functions.

Since `restoreFocus` seems to be the default to use externally, I think it makes sense to keep both public.

test/spec/features/bendpoints/BendpointsSpec.js
```diff
@@ -592,14 +592,16 @@ describe('features/bendpoints', function() {
 
         var bendpointContainer = domQuery('.djs-bendpoints', layer),
             draggerGfx = bendpoints.getSegmentDragger(1, bendpointContainer),
-            draggerVisual = getVisual(draggerGfx),
-            oldBounds = draggerVisual.getBoundingClientRect();
+            draggerVisual = getVisual(draggerGfx);
```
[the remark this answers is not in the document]
jarekdanielak: This was part of your original change. I don't see any effects reverting this, test still passes. Should I revert it just for git history clarity?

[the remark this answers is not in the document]
nikku: Yes, we don't want to have random changes in this PR.

lib/features/popup-menu/PopupMenu.js
jarekdanielak: I don't think this is relevant anymore since this logic was moved to `Canvas`.

lib/features/keyboard/Keyboard.js
```diff
@@ -24,10 +22,11 @@ import {
 var KEYDOWN_EVENT = 'keyboard.keydown',
     KEYUP_EVENT = 'keyboard.keyup';
 
-var HANDLE_MODIFIER_ATTRIBUTE = 'input-handle-modified-keys';
-
 var DEFAULT_PRIORITY = 1000;
 
+var compatMessage = 'Explicit binding of keyboard to an element got removed. For more information, see https://github.com/bpmn-io/diagram-js/issues/661';
```
nikku, line 27:
```suggestion
var compatMessage = 'Keyboard binding is now implicit; explicit binding to an element got removed. For more information, see https://github.com/bpmn-io/diagram-js/issues/661';
```

test/spec/core/CanvasSpec.js
```diff
@@ -68,6 +74,97 @@ describe('Canvas', function() {
...
+        expect(clientRect).to.eql(svg.getBoundingClientRect());
+      }));
+
+
+      it('should not focus on existing document focus', inject(function(canvas, eventBus) {
+
+        // given
+        var inputEl = document.createElement('input');
+
+        document.body.appendChild(inputEl);
```
nikku, line 145: We want to remove such element after the test, I think.

test/spec/features/popup-menu/PopupMenuSpec.js
```diff
@@ -1974,7 +1992,7 @@ describe('features/popup-menu', function() {
 
       var y = documentBounds.height - 40;
 
-      var cursorPosition = { x: documentBounds.left + 100, y: documentBounds.top + y };
+      var cursorPosition = { x: documentBounds.left + 100, y };
```
nikku, line 1995: Does this signal a break in behavior, or is this just a test simplification?

test/spec/core/CanvasSpec.js
```diff
@@ -68,6 +74,100 @@ describe('Canvas', function() {
...
+        eventBus.fire('element.hover', {
+          element: canvas.getRootElement(),
+          gfx: svg
+        });
+
+        // then
+        expect(document.activeElement).to.eql(inputEl);
+
+        // clean up
+        document.body.removeChild(inputEl);
```
nikku, line 163: Has be be cleaned up safely in an _afterEach_ block.

nikku (commented):
Approved :heavy_check_mark:.

Great work on following up @jarekdanielak :rocket:

PR #668, opened 2022-08-18
Revert breaking

nikku:
Reverts what are unpublished (and incomplete) changes in the realm of browser native focus handling / implicit keyboard binding.


philippfromme (approved):
😢

PR #669, opened 2022-08-19
Switch to `babel-plugin-istanbul` for coverage reporting

nikku:
`istanbul-instrumenter-loader` is discontinued. Switching to recommended alternative.


pinussilvestrus (approved):
I like this 👏 CodeCov seems to be happy, too 🙂 

image

PR #674, opened 2022-09-09
Use ES2018 syntax

barmac:
# Breaking Changes

* Migrated to ES2018 syntax. [Read the blog post with details and a migration guide](https://bpmn.io/blog/posts/2022-core-libraries-migrated-to-es2018.html).

---

Requires https://github.com/bpmn-io/bpmn.io/pull/147

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,8 +1,5 @@
 {
-  "extends": "plugin:bpmn-io/es6",
-  "env": {
-    "browser": true
-  },
+  "extends": "plugin:bpmn-io/browser",
   "rules": {
     "no-restricted-imports": [ "error", {
       "name": "inherits",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ _**Note:** Yet to be released changes appear here._
 
 ### Breaking Changes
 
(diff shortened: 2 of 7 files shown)


nikku (commented):
I don't quite understand why we ported the `Canvas` from prototypical inheritance to `class`. Sure it is ES2018, but so are other things.

My feeling is that we should rather keep a consistent style, i.e. introduce `let` and `const` consistently across the code base.

nikku (changes requested):
Let us unbuild the partial prototype -> class migration for the moment as discussed and consider a holistic upgrade (i.e. via [codemod](https://github.com/cpojer/js-codemod/)) at some point.

barmac:
I removed Model migration because it was mostly about `class`, and re-migrated Canvas to showcase const/let and arrow function only.

barmac:
I migrated also CommandStack to showcase more modern syntax, e.x. `for..of` loop: https://github.com/bpmn-io/diagram-js/pull/674/commits/913717a55f3af44971ed0c39f4d4210fa702262e#diff-51be9f9e4dde1107a6fb2a7e3b477a4fac1ecc7828a9fd1cfc3c88aeea6b0403R353

barmac:
Ready for another round.

PR #681, opened 2022-10-25
fix: KeyboardEvent.keyCode is deprecated

kerlon5:
use KeyboardEvent.code instead
it is used to fix [camunda/camunda-modeler/2724](https://github.com/camunda/camunda-modeler/issues/2724)

I'm not sure, do you need an old check by KeyboardEvent.keyCode? Then we can use triple check

### DIFF
--- a/lib/features/dragging/Dragging.js
+++ b/lib/features/dragging/Dragging.js
@@ -25,6 +25,8 @@ import {
   delta as deltaPos
 } from '../../util/PositionUtil';
 
+import { isKey } from '../keyboard/KeyboardUtil';
+
 var DRAG_ACTIVE_CLS = 'djs-drag-active';
 
 
@@ -282,7 +284,7 @@ export default function Dragging(eventBus, canvas, selection, elementRegistry) {
 
   function checkCancel(event) {
 
-    if (event.which === 27) {
+    if (isKey('Escape')) {
       preventDefault(event);
 
       cancel();
--- a/lib/features/keyboard/KeyboardUtil.js
+++ b/lib/features/keyboard/KeyboardUtil.js
@@ -1,14 +1,9 @@
 import { isArray } from 'min-dash';
 
-var KEYCODE_C = 67;
-var KEYCODE_V = 86;
-var KEYCODE_Y = 89;
-var KEYCODE_Z = 90;
-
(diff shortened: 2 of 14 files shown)


kerlon5:
@barmac I can't add "needs review" label to PR

barmac:
That is fine. We have an automation for this :)

barmac:
It's OK, we don't need the flag to review the PR.

lib/features/keyboard/KeyboardUtil.js
```diff
@@ -59,6 +59,10 @@ export function isPaste(event) {
   return isCmd(event) && isKey(KEYS_PASTE, event);
 }
 
+export function isCut(event) {
+  return isCmd(event) && isKey(KEYS_CUT, event);
+}
+
```
barmac, line 65: Neither this nor `isSelectAll` are used anywhere in the code.

barmac (changes requested):
While I believe the general direction of the PR, it does not replace keyCode with supported properties in the whole codebase. Notice that SearchPad is left intact: https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/search-pad/SearchPad.js#L94

lib/features/keyboard/KeyboardUtil.js
kerlon5: removed unsed code and keyCode on all codebase - ready to review

barmac:
I will look into this again this week. Sorry for delay.

lib/features/dragging/Dragging.js
```diff
@@ -282,7 +282,7 @@ export default function Dragging(eventBus, canvas, selection, elementRegistry) {
 
   function checkCancel(event) {
 
-    if (event.which === 27) {
+    if (event.code === 'Escape') {
```
barmac, line 285: This will break the feature for IE entirely: https://caniuse.com/keyboardevent-code

nikku:
@barmac I believe we can ditch IE entirely now. New features we build (cf. https://github.com/bpmn-io/diagram-js/pull/687) don't have the legacy in sight, too.

barmac (approved):
I adjusted the PR and added a changelog update. Ready to merge.

barmac:
So there is actually a polyfill for KeyboardEvent#key which I linked :)

PR #696, opened 2022-11-16
fix(popup): `_getEntry` to search in headerEntries

### LINKED ISSUE #695: Popup menu trigger action doesn't work for header entries
__Describe the Bug__

Popup menu trigger action doesn't work for header entries. This was introduced with the new popup menu (#686) because [`popupMenu._current.entries`](https://github.com/bpmn-io/diagram-js/blob/a68b7c6fa272c6cbe250f9255c200c0684fc32ef/lib/features/popup-menu/PopupMenu.js#L168) now only represents body entries. 

__Steps to Reproduce__

1. try to trigger action of a header entry (eg `popupMenu.trigger(globalEvent(headerEntry, { x: 0, y: 0 }));` or see [bpmn-js test case](https://github.com/bpmn-io/bpmn-js/blob/60ad7d4696baf7fb839704394b332cf424ff4824/test/spec/features/popup-menu/ReplaceMenuProviderSpec.js#L503))

__Expected Behavior__

`popupMenu.trigger` should continue to work as before: allowing triggering of all entries.

__Environment__

 - Library version: 11.0.0-aplha.0

smbea:
Closes #695

### DIFF
--- a/lib/features/popup-menu/PopupMenu.js
+++ b/lib/features/popup-menu/PopupMenu.js
@@ -501,7 +501,7 @@ PopupMenu.prototype.trigger = function(event) {
  */
 PopupMenu.prototype._getEntry = function(entryId) {
 
-  var entry = this._current.entries[entryId];
+  var entry = this._current.entries[entryId] || this._current.headerEntries[entryId];
 
 
   if (!entry) {
--- a/test/spec/features/popup-menu/PopupMenuSpec.js
+++ b/test/spec/features/popup-menu/PopupMenuSpec.js
@@ -636,18 +636,33 @@ describe('features/popup-menu', function() {
               }
             }
           ];
+        },
+        getHeaderEntries: function() {
+          return [
+            {
+              id: '3',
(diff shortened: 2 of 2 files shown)


nikku:
Some observations here: The old popup menu [actually used `PopupMenu#trigger`](https://github.com/bpmn-io/diagram-js/commit/ce3170be9021cc1b489c482ccb5849087c50ec8f#diff-20749ae7944fbc3d18a7ab06c462b53d0507502d6271eaa0e354394580b233c3L391) to trigger the event behavior. The new one does not do that anymore, effectively working around the public API. 

Shall we adjust the new menu to use `PopupMenu#trigger`?

nikku:
GitHub messed up the deep link, let me inline it: 

image

The old popup menu actually used `PopupMenu#trigger` internally, on entry click. The new one does not do that. Yes, the API still exists, but it is not used, under the hood. 

My suggestion is now to consider restoring that behavior (if it makes sense).

nikku (approved):

PR #708, opened 2022-11-22
Make keyboard friendly enough

### LINKED ISSUE #707: Make keyboard more friendly
__What should we do?__

Make keyboard more friendly in event handling. It shouldn't force handle events.

**Current behaviour:**

* It intercepts during capture phase
   * Handles events even if component handles them
* Does not consider if element has `tabIndex`
* Input is not considered only in some cases: https://github.com/bpmn-io/diagram-js/blob/5f4223509f838e894a8ae75899b46ae93ca2a8cc/lib/features/keyboard/Keyboard.js#L112

It should allow external component to handle their events nicely: if component prevents default, keyboard shouldn't handle it

__Why should we do it?__

Allow easier handling of events by components.

nikku:
Ensure we play nice with other UI components up in the component tree; rather than handling keys on capture, handle events on bubble. Ignore events that got their propagation stopped or their default prevented.

A result of this PR is that components downwards in the component tree can use standard DOM facilities to prevent the event from being handled by keyboard:

```
(e)
 |
 * <Some Component> (e.stopPropagation())
 
 * <Keyboard>
```
-----

Closes #707

### DIFF
--- a/lib/features/keyboard/Keyboard.js
+++ b/lib/features/keyboard/Keyboard.js
@@ -24,7 +24,7 @@ var DEFAULT_PRIORITY = 1000;
 
 /**
  * A keyboard abstraction that may be activated and
- * deactivated by users at will, consuming key events
+ * deactivated by users at will, consuming global key events
  * and triggering diagram actions.
  *
  * For keys pressed down, keyboard fires `keyboard.keydown` event.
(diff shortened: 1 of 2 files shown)


nikku:
I did integration test this against bpmn-js and found https://github.com/bpmn-io/diagram-js/pull/709 (unrelated to this change).

barmac:
Great improvement 👏

PR #714, opened 2022-11-25
Fix upgrade issues

nikku:
Testing the diagram-js@11.1.0 integration into bpmn-js I found two quirks that are worth investigating:

* We broke the semantics of the `djs-popup` class, unnecessarily => proposal to revert this via https://github.com/bpmn-io/diagram-js/commit/71f1f1c094fcb72a943c7a2fa8eb568a5487e72a
* We apply inconsistent padding / margin leading to broken UI in search/header-less popups => Fixed via https://github.com/bpmn-io/diagram-js/pull/714/commits/c28fadf21b06613a8b37ec72df104d162fa69ff3

----

#### Screens showcasing UI after padding/margin fixes

screenshot 6a40eh
screenshot iUAln8
screenshot zTDADd

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -555,13 +555,6 @@ marker.djs-dragger tspan {
   outline: none;
 }
 
-.djs-popup_header {
-  display: flex;
-  align-items: stretch;
-  line-height: 20px;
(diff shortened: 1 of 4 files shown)


nikku:
Thanks for the quick review. :star:

PR #718, opened 2022-12-07
feat(popup-menu): fire `opened`/`closed` events

smbea:
Related to https://github.com/camunda/camunda-modeler/issues/3342

### DIFF
--- a/lib/features/popup-menu/PopupMenu.js
+++ b/lib/features/popup-menu/PopupMenu.js
@@ -122,6 +122,8 @@ PopupMenu.prototype._render = function() {
         entries=${ entriesArray }
         headerEntries=${ headerEntriesArray }
         scale=${ scale }
+        onOpened=${ this._onOpened.bind(this) }
+        onClosed=${ this._onClosed.bind(this) }
         ...${{ ...options }}
       />
     `,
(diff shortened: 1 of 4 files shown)


lib/features/popup-menu/PopupMenu.js
```diff
@@ -176,6 +176,8 @@ PopupMenu.prototype.open = function(element, providerId, position, options) {
   this._bindAutoClose();
 
   this._render();
+
+  this._emit('opened');
```
nikku, line 180: Emit when actually opened (i.e. mounted by react component)?

lib/features/popup-menu/PopupMenu.js
```diff
@@ -122,6 +122,7 @@ PopupMenu.prototype._render = function() {
         entries=${ entriesArray }
         headerEntries=${ headerEntriesArray }
         scale=${ scale }
+        emitEvent=${ this._emit.bind(this) }
```
nikku, line 125: Please expose standard `onOpen` and `onClose` handlers.

nikku: Is there a reason why we emit closed throught the PREACT component? We should be fine with emitting it directly from the popup menu.

test/spec/features/popup-menu/PopupMenuComponentSpec.js
```diff
@@ -6,9 +6,7 @@ import {
 } from 'lib/ui';
 
 import {
-  bootstrapDiagram,
-  insertCSS,
-  inject
+  insertCSS
```
nikku, line 9: Please add tests that verify the API in the API spec (`PopupMenuSpec`). This component may be dead, but the contract we setup shall remain.

lib/features/popup-menu/PopupMenu.js
nikku: Whatever you decide to go for, let's adhere to a common components convention: Props that are passed into a component are called `on{NAME}`. If you go the generic _emit any event_ route, call the thing `onEvent` or `onLifeCycleEvent`. I'd advocate for a specific handler though.

nikku (approved):

PR #719, opened 2022-12-07
contextPad: allow keyboard trigger

smbea:
Related to https://github.com/bpmn-io/bpmn-js/pull/1785, see https://github.com/bpmn-io/bpmn-js/pull/1785#issuecomment-1337081878

### DIFF
--- a/test/spec/features/context-pad/ContextPadSpec.js
+++ b/test/spec/features/context-pad/ContextPadSpec.js
@@ -79,7 +79,7 @@ describe('features/context-pad', function() {
       // then
       expect(function() {
         contextPad.registerProvider(provider);
-      }).to.not.throw;
+      }).not.to.throw;
     }));
 
 
@@ -205,7 +205,7 @@ describe('features/context-pad', function() {
         var entries = contextPad.getEntries([]);
 
         // then
-        expect(entries.entryA).to.not.exist;
+        expect(entries.entryA).not.to.exist;
       }));
 
 
@@ -229,7 +229,7 @@ describe('features/context-pad', function() {
           var entries = contextPad.getEntries([]);
 
           // then
-          expect(entries.entryA).to.not.exist;
+          expect(entries.entryA).not.to.exist;
         }));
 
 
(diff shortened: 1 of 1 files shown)


lib/features/context-pad/ContextPad.js
```diff
@@ -239,6 +239,35 @@ ContextPad.prototype.trigger = function(action, event, autoActivate) {
...
 
+/**
+ * Trigger context pad action with entry id.
+ *
+ * @param {String} entryId
+ * @param {string} action
+ * @param {Event} event
+ * @param {boolean} [autoActivate=false]
+ */
+ContextPad.prototype.triggerWithId = function(entryId, action, event, autoActivate) {
```
nikku, line 250: Let#s consolditate this. Can'T we make this a low level function, used by `trigger` (instead of duplicating)?

nikku: > But I think it should still be a public API

Agreed. Just good to re-use things where possible, to prevent bugs in what should be shared behavior.

lib/features/overlays/Overlays.js
```diff
@@ -294,6 +294,9 @@ Overlays.prototype.remove = function(filter) {
 
 };
 
+Overlays.prototype.isShown = function() {
```
nikku, line 297: Please add a test case that verifies the existence of this API.

nikku (changes requested):

lib/features/overlays/Overlays.js
nikku: What I was asking for is covering the `Overlays#isShown` :see_no_evil: (`OverlaysSpec` or something). Will add the test case now. :+1:

nikku: Added via https://github.com/bpmn-io/diagram-js/pull/719/commits/07e28f0501d6d131a2fa885ccea8b3776c57e059.

nikku: Added another test to verify the newly added `ContextPad` API, which I renamed to `ContextPad#triggerEntry` (for clarity).

nikku: We did not return a value anymore from `trigger`. Fixed + aadded test case to it.

nikku (approved):

PR #723, opened 2022-12-12
fix(popup-menu): increase z-index

smbea:
Related to https://github.com/camunda/camunda-modeler/issues/3349

[Link to screenshots of how it looks in desktop modeler](https://github.com/camunda/camunda-modeler/issues/3349#issuecomment-1347225562)

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -521,7 +521,7 @@ marker.djs-dragger tspan {
   height: 100vh;
   top: 0;
   left: 0;
-  z-index: 200;
+  z-index: 2000;
   line-height: 1;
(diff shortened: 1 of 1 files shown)


assets/diagram-js.css
```diff
@@ -521,7 +521,7 @@ marker.djs-dragger tspan {
   height: 100vh;
   top: 0;
   left: 0;
-  z-index: 200;
+  z-index: 2000;
```
nikku, line 524: Why 2000 over 200 over 1377? :slightly_smiling_face: 

`z-index` is already 200 times higher than the base. Overrides are possible within applications.

PR #728, opened 2023-01-05
Render popup menu in `.djs-container`

smbea:
Related to https://github.com/bpmn-io/bpmn-js/issues/1795

See comment https://github.com/bpmn-io/bpmn-js/issues/1795#issuecomment-1372016215.

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -524,8 +524,6 @@ marker.djs-dragger tspan {
   z-index: 200;
   line-height: 1;
   font-family: "IBM Plex Sans", sans-serif;
-  border: solid 1px var(--popup-border-color);
-  border-radius: 2px;
 }
 
 .djs-popup {


nikku:
Regarding https://github.com/bpmn-io/bpmn-js/issues/1795#issuecomment-1372016215: Any reason we don't simply attach it to `djs-container` (again)? What is the use-case to make it configurable?

lib/features/popup-menu/PopupMenu.js
```diff
@@ -237,11 +237,7 @@ PopupMenu.prototype._onClosed = function() {
 
 PopupMenu.prototype._createContainer = function(config) {
 
-  let parent = config && config.parent || 'body';
-
-  if (typeof parent === 'string') {
-    parent = document.querySelector(parent);
-  }
+  let parent = document.querySelector('.djs-container');
```
nikku, line 240: We'd need to retrieve it from the [diagram container](https://github.com/bpmn-io/diagram-js/blob/develop/lib/core/Canvas.js#L491) (cf. original implementation) or we risk finding a adjacent diagram container, i.e. if multiple diagram-js instances are embedded on the same page.

nikku:
Added a [test case](https://github.com/bpmn-io/diagram-js/pull/728/files#diff-7dc68f128643d145cf2a06ab4ce54a7cc10388034805b1f6cbe730303853583dR274).

nikku (approved):

PR #731, opened 2023-01-16
feat(popup-menu): allow entries to be draggable

smbea:
Related to https://github.com/bpmn-io/bpmn-js/pull/1802 -> https://github.com/bpmn-io/bpmn-js/pull/1802#issuecomment-1383256080

### DIFF
--- a/lib/features/popup-menu/PopupMenuComponent.js
+++ b/lib/features/popup-menu/PopupMenuComponent.js
@@ -232,7 +232,7 @@ export default function PopupMenuComponent(props) {
             entries=${ entries }
             selectedEntry=${ selectedEntry }
             setSelectedEntry=${ setSelectedEntry }
-            onClick=${ onSelect }
+            onAction=${ onSelect }
           />
         </div>
         ${ entries.length === 0 && html`
--- a/lib/features/popup-menu/PopupMenuItem.js
+++ b/lib/features/popup-menu/PopupMenuItem.js
@@ -24,8 +24,6 @@ export default function PopupMenuItem(props) {
     onAction
   } = props;
 
-  const draggable = entry.action && entry.action.dragstart;
-
   return html`
     <li
       class=${ classNames('entry', { selected }) }
@@ -35,7 +33,7 @@ export default function PopupMenuItem(props) {
       onMouseEnter=${ onMouseEnter }
       onMouseLeave=${ onMouseLeave }
       onDragStart=${ (event) => onAction(event, entry, 'dragstart') }
-      draggable=${ draggable }
+      draggable=${ true }
     >
       <div class="djs-popup-entry-content">
         <span
--- a/lib/features/popup-menu/PopupMenuItem.js
+++ b/lib/features/popup-menu/PopupMenuItem.js
@@ -13,7 +13,7 @@ import {
  * @param {boolean} selected
  * @param {function} onMouseEnter
  * @param {function} onMouseLeave
- * @param {function} onClick
+ * @param {function} onAction
  */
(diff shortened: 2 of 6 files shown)


lib/features/popup-menu/PopupMenuComponent.js
```diff
@@ -232,7 +234,8 @@ export default function PopupMenuComponent(props) {
             entries=${ entries }
             selectedEntry=${ selectedEntry }
             setSelectedEntry=${ setSelectedEntry }
-            onClick=${ onSelect }
+            onSelect=${ onSelect }
+            draggable=${ draggable }
```
nikku, line 238: Where does this magic "draggable" come from? Do we deduce "element is draggable" from a specific popup menu configuration? Or from a `dragstart` handler being provided with the entry?

nikku: And what is the up or downside of either approach?

(We could likely take inspiration from the context pad and implement exactly what we got there, it served us well).

nikku: My suggestion would be: Compute the `draggable` property from the presence of the `dragstart` handler in an entries `action` member.

It does not make sense to drag an entry with 

```
{
  label: 'FOO',
  action: (event, ...) => { ... }
}
```

You have to have it `draggable` with the following configuration though: 

```
{
  label: 'FOO',
  action: {
    dragstart: (event, ...) => { ... }
  }
}
```

nikku: Feels odd to need to pass it on (explicitly).

nikku:
@smbea I've extracted the "complex" test case into an integration test. It should be easier now to verify if the stuff works end-to-end (and how it works). 

Going back to https://github.com/bpmn-io/diagram-js/pull/731#discussion_r1071123490, did you validate the actual feel? Any odd behaviors you found?

I do like that `draggable` is set implicitly.

nikku:
Is there a concrete reason to not mirror the [ContextPad behavior](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/context-pad/ContextPad.js#L252)? 

The end result should be a robust, predictable behavior for the user.

lib/features/popup-menu/PopupMenuItem.js
```diff
@@ -13,25 +13,29 @@ import {
...
   const {
     entry,
     selected,
     onMouseEnter,
     onMouseLeave,
-    onClick
+    onSelect
   } = props;
 
+  const draggable = entry.action && entry.action.dragstart ? true : false;
```
barmac, line 27:
```suggestion
  const draggable = entry.action && entry.action.dragstart
```

lib/features/popup-menu/PopupMenuItem.js
```diff
@@ -13,25 +13,29 @@ import {
  * @param {boolean} selected
  * @param {function} onMouseEnter
  * @param {function} onMouseLeave
- * @param {function} onClick
+ * @param {function} onSelect
```
barmac, line 16: To me, it feels a bit confusing to talk about `onSelect` outside of select event context. Why don't we call it `onAction` so that it's the same name as in `entry.action`?

barmac:
Looks good overall. I left a couple of comments which you can consider to implement but IMO the PR is mergable as is.

nikku:
UI wise always making the entries draggable and later canceling the default is the more stable behavior; it prevents ugly artifacts (text selection) as seen in the capture below: 

capture KlQ7Qa_optimized

I've updated the PR to that behavior (which we also have in `context-pad`, by the way).

PR #733, opened 2023-01-17
feat(create): allow create with keyboard

smbea:
Related to https://github.com/bpmn-io/bpmn-js-connectors-extension/issues/54 and https://github.com/bpmn-io/bpmn-js/issues/1801#issuecomment-1383708152

### DIFF
--- a/lib/features/create/Create.js
+++ b/lib/features/create/Create.js
@@ -34,7 +34,8 @@ export default function Create(
     dragging,
     eventBus,
     modeling,
-    rules
+    rules,
+    mouse
 ) {
 
   // rules //////////
@@ -307,6 +308,10 @@ export default function Create(
       });
     });
 
+    if (event instanceof KeyboardEvent) {
+      event = mouse.getLastMoveEvent();
+    }
+
     dragging.init(event, PREFIX, {
       cursor: 'grabbing',
       autoActivate: true,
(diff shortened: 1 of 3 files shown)


lib/features/create/Create.js
```diff
@@ -307,6 +308,10 @@ export default function Create(
       });
     });
 
+    if (event instanceof KeyboardEvent) {
```
nikku, line 311: There are two options to go about this: 

Make `Create.start` work out-of-the-box, implicitly using the mouse utility _or_ explicitly invoke `Create.start` with a mouse event (that you got from wherever).

While I like your solution we did it differently in the past, cf. https://github.com/bpmn-io/diagram-js/blob/9df88480eca290ba16af2f36b5a4a656583e10df/lib/features/copy-paste/CopyPaste.js#L242.

Did you consider the alternative?

nikku: Let us consider it like this:

What value do I get from passing in a `KeyEvent` over a `MouseEvent` over `null` into `create.start(...)`? If there is none then I'd advocate to make it possible to call `create.start` without an event, defaulting to the last mouse move event, if provided.

If there is one, let's ensure that we don't break that.

nikku: > But I can also see how it may look random in the middle of create.

We could make the dependency to `mouse` an optional one (`injector.get('mouse', false)`) and clearly document that the last good known mouse position is chosen, if no event is provided.

nikku: > I don't have a super strong opinion: if we would rather keep this on a need-to-use basis locally, i'm also happy would that.

If this does not involve much work I'd favor the local solution, I think.

PR #740, opened 2023-01-19
fix: align with `alignment[type]=0`

### LINKED ISSUE #739: Alignment does not work with zero coord
### Describe the Bug

An element with `x=0` or `y=0` may fail to run align command.

### Steps to Reproduce

- Create a shape with `x=0`
- Run align left command, with shapes that `x>0`
- Nothing happened.

see https://codesandbox.io/s/diagram-js-align-bug-myob79?file=/src/index.js

### Expected Behavior

Align as expected.

### Environment

 - Browser: Microsoft Edge 108
 - OS: macOS 13.0.1 (22A400)
 - Library version: 11.5.0

YuJianghao:
fix #739

### DIFF
--- a/lib/features/modeling/cmd/AlignElementsHandler.js
+++ b/lib/features/modeling/cmd/AlignElementsHandler.js
@@ -1,4 +1,4 @@
-import { forEach } from 'min-dash';
+import { forEach, isNumber } from 'min-dash';
 
 /**
  * A handler that align elements in a certain way.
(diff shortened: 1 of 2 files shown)


lib/features/modeling/cmd/AlignElementsHandler.js
```diff
@@ -25,22 +25,22 @@ AlignElements.prototype.preExecute = function(context) {
       y: 0
     };
 
-    if (alignment.left) {
+    if (isNumber(alignment.left)) {
```
YuJianghao, line 28: Can `alignment[type]` be string? Cause Javascript also support `'1' + 1 = 2`.

nikku: Let us assume it cannot. We could guard against this by using `isDefined(alignment.left)` instead of the number check.

nikku:
Great stuff. Could you also [add a test case](https://github.com/bpmn-io/diagram-js/blob/develop/test/spec/features/align-elements/AlignElementsSpec.js) to verify alignment around `0`?

nikku:
Great stuff. Thanks for your contribution.

PR #741, opened 2023-01-24
feat(palette): allow trigger with entry id

smbea:
In order to support https://github.com/bpmn-io/bpmn-js/issues/1814; similar to how we did for context pad https://github.com/bpmn-io/diagram-js/pull/719

I added some general test cases for `#trigger` similar to the ones we have for context pad since we didn't have these for palette

### DIFF
--- a/lib/features/palette/Palette.js
+++ b/lib/features/palette/Palette.js
@@ -285,9 +285,7 @@ Palette.prototype._update = function() {
  * @param  {Event} event
  */
 Palette.prototype.trigger = function(action, event, autoActivate) {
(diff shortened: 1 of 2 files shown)


nikku (approved):
Great stuff.

barmac (approved):

nikku:
You got mine because of 

> I added some general test cases for #trigger similar to the ones we have for context pad since we didn't have these for palette

Not sure why you got @barmac's :wink:

barmac:
This makes sense.

PR #750, opened 2023-02-18
refactor: replace npm module `css.escape` with native API

thetric:
Related to https://github.com/bpmn-io/bpmn-js/issues/1850


nikku (commented):
Awesome!

PR #755, opened 2023-03-03
feat: support select after replacement

nikku:
This mirrors existing patterns (cf. auto-place).

Per default replaced elements will be selected, unless explicitly disabled via hint.

----

Idea is to get interaction related logic, where a shared concern, out of downstream libraries (properties panel [element template removal](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/663003e48cdb19942f9d7293e7269562eb8aae97/src/provider/element-templates/util/templateUtil.js#L33), [bpmn replace](https://github.com/bpmn-io/bpmn-js/blob/develop/lib/features/replace/BpmnReplace.js#L290)).

### DIFF
--- a/lib/features/replace/Replace.js
+++ b/lib/features/replace/Replace.js
@@ -13,23 +13,23 @@ var round = Math.round;
  *
  * @param {Modeling} modeling
  */
-export default function Replace(modeling) {
-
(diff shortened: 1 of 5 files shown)


philippfromme (approved):
Cool! ✅

Issue #758, opened 2023-03-10
EventCallback type does not allow additional parameters


chaptergy:
### Describe the Bug

The new EventCallback type defines as the callback only having the single `event` parameter. However when firing the event, you can add an arbitrary number of additional parameters.

### Steps to Reproduce

Use the EventBus.on method within a typescript context with additonal parameters provided by the event. The TS-compiler complains about there beint to many arguments.

### Expected Behavior

I can define the callback function with additional parameters without a problem

### Environment

Please complete the following information:

 - Browser: not relevant
 - OS: not relevant
 - Library version: 11.11.0

PR #759, opened 2023-03-10
Adds additional parameters to EventCallback type

chaptergy:
Currently the type defines, that the EventCallback can only receive a single parameter `event`. However when firing an event, an arbitrary amount of additional parameters can be passed to the callback.

Closes https://github.com/bpmn-io/diagram-js/issues/758.

### DIFF
--- a/lib/core/EventBus.d.ts
+++ b/lib/core/EventBus.d.ts
@@ -6,7 +6,7 @@ export type Event = {
   defaultPrevented: boolean;
 };
 
-export type EventCallback<T extends Event> = (event: T) => any | void;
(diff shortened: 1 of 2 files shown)


nikku:
Thanks. This makes a lot of sense.

Could you add a [usage example here](https://github.com/bpmn-io/diagram-js/blob/develop/lib/core/EventBus.spec.ts#L47) to verify your intended API use is covered?

nikku (changes requested):
We'd want to add a test case for this, cf. https://github.com/bpmn-io/diagram-js/pull/759#issuecomment-1474076765.

nikku (approved):
Thanks.

PR #763, opened 2023-03-23
chore: remove internal visual grid

nikku:
Preparation to move visual grid out of diagram-js, to live a more prosper live in a [dedicated extension](https://github.com/bpmn-io/diagram-js-grid).


smbea (approved):
Nice clean up 🧹

PR #772, opened 2023-04-05
Emit `popupMenu/palette/contextPad.trigger` events

### LINKED ISSUE #771: Emit popupMenu/palette/contextPad `trigger` events
### What should we do?

We should fire eventBus events when popup menu, palette or context pad actions are triggered.

- `popupMenu.trigger`
- `palette.trigger`
- `contextPad.trigger`

### Why should we do it?

This will facilitate tracking, see https://github.com/bpmn-io/internal-docs/issues/746#issuecomment-1495885885

smbea:
Closes #771

### DIFF
--- a/lib/features/context-pad/ContextPad.js
+++ b/lib/features/context-pad/ContextPad.js
@@ -259,6 +259,10 @@ ContextPad.prototype.triggerEntry = function(entryId, action, event, autoActivat
 
   var handler = entry.action;
 
+  if (this._eventBus.fire('contextPad.trigger', { entry, event }) === false) {
+    return;
+  }
+
   // simple action (via callback function)
   if (isFunction(handler)) {
     if (action === 'click') {
--- a/test/spec/features/context-pad/ContextPadSpec.js
+++ b/test/spec/features/context-pad/ContextPadSpec.js
@@ -964,6 +964,39 @@ describe('features/context-pad', function() {
   });
 
 
+  describe('event integration', function() {
+
+    beforeEach(bootstrapDiagram({ modules: [ contextPadModule, providerModule ] }));
+
+    it('should fire "contextPad.trigger"', inject(function(canvas, contextPad, eventBus) {
+
+      // given
+      var shape = canvas.addShape({ id: 's1', width: 100, height: 100, x: 10, y: 10 });
+      var triggerSpy = sinon.spy();
+
+      eventBus.on('contextPad.trigger', triggerSpy);
+
+      contextPad.open(shape);
+
+      var pad = contextPad.getPad(shape),
+          html = pad.html,
+          target = domQuery('[data-action="action.c"]', html);
+
+      var event = globalEvent(target, { x: 0, y: 0 });
+
+      // when
+      contextPad.trigger('click', event);
+
+      // then
+      const entry = contextPad._current.entries['action.c'];
+
+      expect(triggerSpy).to.have.been.calledOnce;
+      expect(triggerSpy.getCall(0).args[1]).to.eql({ entry, event });
+    }));
+
+  });
+
+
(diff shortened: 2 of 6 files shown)


lib/features/context-pad/ContextPad.js
```diff
@@ -259,6 +259,8 @@ ContextPad.prototype.triggerEntry = function(entryId, action, event, autoActivat
 
   var handler = entry.action;
 
+  this._eventBus.fire('contextPad.trigger', { event });
```
nikku, line 262: Shall we allow intercepting of trigger?
```suggestion
  if (this._eventBus.fire('contextPad.trigger', { action, event }) === false) {
    return;
  };
```

test/spec/features/context-pad/ContextPadSpec.js
```diff
@@ -794,6 +802,7 @@ describe('features/context-pad', function() {
       contextPad.trigger('click', event);
 
       // then
+      expect(triggerSpy).to.have.been.calledOnce;
```
nikku, line 805: Let's assert what it has been called with; potentially in a separate spec ("event integration").

nikku (changes requested):

lib/features/context-pad/ContextPad.js
smbea: Makes sense. I added `entry` and `event` as parameters because we can get the label for the target element via `entry` (user triggered "Start Event" entry); and the type of `event` allows us infer the trigger source: drag vs click vs keyboard

nikku: Why not expose "action", too? Given that this is public API (cf. keyboard integration) the authoritive truth should be the `action` parameter, not the event type, or am I wrong?

smbea: We could do that too. `action` is one of "click" or "drag". Triggering via keyboard, acts as a click and therefore the `action` is click too. But by checking the event type we can actually differentiate it.

nikku: > We could instead, send the explicit action already based on the event type. This way, listeners to it wouldn't have to work that out themselves.

Or add `keyboard` as an action, that equals `click` in behavior.

smbea: Getting this information via the event would be something like this:
```javascript
if(event instanceOf PointerEvent) { type = "click"}
else if(event instanceOf DragEvent) {type = "drag"}
else if(event instanceOf KeyboardEvent) {type = "keyboard"}
```

Adding an `action` for keyboard means changing the providers in bpmn-js and bpmn-js-create-append-anything.
If we do want to sned `action` explicitly, it would be easier to infer that in diagram-js from the event and send it.

test/spec/features/palette/PaletteSpec.js
```diff
@@ -798,6 +798,53 @@ describe('features/palette', function() {
...
+      }
+    };
+
+    beforeEach(bootstrapDiagram({ modules: [ paletteModule ] }));
+
+    beforeEach(inject(function(palette) {
+      palette.registerProvider(800, paletteProvider);
+    }));
+
+    afterEach(function() {
```
nikku, line 825: I personally like to have _less_ stuff like this outside of a spec. Otherwise I have to context switch too much (scroll to see what "triggerSpy" is and so forth). `triggerSpy` could be setup inside the test (I think), and then we'd not need this.

smbea: Gotcha, I moved it to inside the spec.

lib/features/context-pad/ContextPad.js
smbea: Let's rely on `event` for now then

PR #779, opened 2023-04-20
Fix issues for clients which have enabled `verbatimModuleSyntax`

boris-petrov:


nikku:
Could you elaborate what this accomplishes? We [generate type definitions](https://unpkg.com/browse/diagram-js@12.0.0/lib/model/Types.d.ts) and there should be no need to touch or consume `Types.ts` at all.

boris-petrov:
Because I get a bunch of:

```
node_modules/diagram-js/lib/model/Types.ts:1:10 - error TS1484: 'Point' is a type and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled.

1 import { Point } from '../util/Types';
           ~~~~~
```

When I import the type-definitions in my project. Where am I supposed to change that?

nikku:
Never mind my previous comment. I've just recognized that the types generated _are_ actually broken: 

image

PR #785, opened 2023-05-08
fix(popupMenu): include entry id in `popupMenu.trigger`

smbea:
Related to https://github.com/bpmn-io/bpmn-js-tracking/issues/21


nikku (approved):
Good find.

PR #790, opened 2023-06-02
feat(popupMenu): allow to provide html for entries

smbea:
Related to https://github.com/bpmn-io/bpmn-js/issues/1918

### DIFF
--- a/lib/features/popup-menu/PopupMenuItem.js
+++ b/lib/features/popup-menu/PopupMenuItem.js
@@ -43,9 +43,8 @@ export default function PopupMenuItem(props) {
         <span
           class=${ classNames('djs-popup-entry-name', entry.className) }
         >
-          ${ entry.imageUrl ? html`
-            <img class="djs-popup-entry-icon" src=${ entry.imageUrl } alt="" />
-          ` : null }
+          ${(entry.imageUrl && html`<img class="djs-popup-entry-icon" src=${ entry.imageUrl } alt="" />`) ||
+            (entry.imageHtml && html`<div class="djs-popup-entry-icon" dangerouslySetInnerHTML=${ { __html: entry.imageHtml } } />`)}
 
           ${ entry.label ? html`
             <span class="djs-popup-label">
--- a/test/spec/features/popup-menu/PopupMenuSpec.js
+++ b/test/spec/features/popup-menu/PopupMenuSpec.js
@@ -1107,6 +1107,32 @@ describe('features/popup-menu', function() {
     }));
 
 
+    it('should add icon as html if specified', inject(function(popupMenu) {
+
+      // given
+      var testMenuProvider = {
+        getEntries: function() {
+          return [
+            {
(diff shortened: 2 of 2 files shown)


barmac:
I'm wondering whether `html` is correct name if the feature is about icon html only. So we still cannot provide HTML for the entire entry as it's then followed by label (in contrast with https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/context-pad/ContextPad.js#L320). Perhaps `iconHtml` could be less confusing here.

Also, do we consider anything passed to the popup menu as trusted? On one hand we could try to enforce some form of sanitization (e.g. we only accept SVG and IMG), but on the other popup menu is quite deeply integrated into the tool and it's implementor's responsibility to make sure no user-controlled input is passed directly to the menu. So we only need to make it clear that the `html/iconHtml` property is not sanitized and make sure that our templates integrations use `imageUrl` or a sanitized HTML only.

barmac:
So the scenario I am thinking about is:
1. I create an element template containing malicious code
2. Somebody loads the template in C8
3. Their sensitive data (e.g. credentials) are sent to my server

smbea:
> Perhaps iconHtml could be less confusing here.

Might make sense since in popup menu the icon is just one element of the entry.

Regarding sanitisation: true it's vulnerable as is. Do we have a current stance on this types of situations already?

barmac:
Check out the security guide: https://github.com/bpmn-io/internal-docs/blob/master/development/SECURITY_GUIDE.md#threat-model
Basically, we trust application code and plugins, but any user-provided content should be sanitized (XML, element templates, user input). IMO from diagram-js perspective the code provided to `iconHtml/html` is application code thus trusted. It is the templates integration's responsibility to sanitize template content.

smbea:
Yeah, makes sense. So what would mean we should check this when validating the templates upon loading.

nikku:
I think we should simply no support loading element template icons as HTML. Why would we want to do it? As you mentioned @barmac this is for controlled, "internal use". Hence I'd be fine to:

* Load element template icons as image url => This is what it is!
* Give integrators the freedom to load their existing icons as html. I think `imageHtml` is the right name in that case, given that we already support `imageUrl`

If we were to start on a green field the better name would probably have been `icon*`.

nikku:
> @nikku You lost me a little with "we should simply no support loading element template icons as HTML" and "give integrators the freedom to load their existing icons as html". Does this mean we don't officially support imageHtml in element templates?

Short answer: YES, we do not support `imageHTML` for element templates.

Long answer: We have the choice how to integrate element templates into the core. From the element template point of view a user provides image content (URL encoded). it is about us creators of the element template framework to decide which capabilities we want to provide. I advocate to provide reasonable features that do not put our users at risk. Hence, no HTML injection, site-loaded by element template descriptors.

nikku (approved):

nikku:
Good stuff!

PR #797, opened 2023-08-21
fix(canvas): `scrollToElement` accepts root elements

marstamm:
related to https://github.com/camunda/camunda-modeler/issues/3758

Recording 2023-08-21 at 14 08 54


philippfromme (approved):
✅

nikku (approved):

PR #806, opened 2023-10-03
feat(zoom-scroll): add zoom by cmd + wheel on mac

webkadiz:
This PR adds support for zooming using cmd on the mac platform

It is clear that zooming with ctrl + wheel is most likely a side effect, since event.ctrlKey is also a directive for pinch zoom. but since there is a feature, why not create a more pleasant UX on Mac

### DIFF
--- a/lib/navigation/zoomscroll/ZoomScroll.js
+++ b/lib/navigation/zoomscroll/ZoomScroll.js
@@ -13,6 +13,10 @@ import {
 } from '../../util/Math';
 
 import {
+  isMac
+} from '../../util/Platform';
+
+import {
   bind
 } from 'min-dash';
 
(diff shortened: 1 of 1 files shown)


barmac:
Hi,

Thanks for opening this PR. 

I've done some tests in the other applications to confirm that we follow users' expectations. Below are the results.

| application | zoom handling |
| - | - |
| GIMP | supports zoom with CMD |
| tldraw | supports zoom with CMD |
| figma | supports zoom with CMD |
| google maps | zoom without CMD |
| excalidraw | supports zoom with CMD |

Based on above, I believe we should support CMD as an alternative key to trigger zoom scrolling. Thank you for implementing this!

barmac (approved):

PR #826, opened 2023-11-09
Connection selection outline + context pad position

smbea:
Related to https://github.com/camunda/product-hub/issues/1653#issuecomment-1803357533

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -121,8 +121,13 @@
   stroke: var(--element-selected-outline-stroke-color);
 }
 
+.djs-connection.selected .djs-outline {
+  display: none;
+}
+
 .djs-multi-select .djs-element.selected .djs-outline {
   stroke: var(--element-selected-outline-secondary-stroke-color);
+  display: block;
 }
 
 .djs-shape.connect-ok .djs-visual > :nth-child(1) {
--- a/lib/features/context-pad/ContextPad.js
+++ b/lib/features/context-pad/ContextPad.js
@@ -23,6 +23,9 @@ import {
   escapeCSS
 } from '../../util/EscapeUtil';
 
+import { isConnection } from '../../util/ModelUtil';
+
+
 /**
  * @typedef {import('../../model/Types').Element} Element
  *
@@ -507,11 +510,19 @@ ContextPad.prototype.isShown = function() {
 /**
  * Get contex pad position.
  *
+ * If target is a connection, the context pad will be placed according to the
+ * connection's last waypoint.
+ *
+ * If multiple targets, the context pad will be placed according to the bounding
+ * box containing all targets.
+ *
  * @param {ContextPadTarget} target
  *
  * @return {Rect}
  */
 ContextPad.prototype._getPosition = function(target) {
+
+  target = isConnection(target) ? getLastWaypoint(target) : target;
 
   var elements = isArray(target) ? target : [ target ];
   var bBox = getBBox(elements);
@@ -545,4 +556,8 @@ function addClasses(element, classNames) {
  */
 function includes(array, item) {
   return array.indexOf(item) !== -1;
+}
+
+function getLastWaypoint(connection) {
+  return connection.waypoints[connection.waypoints.length - 1];
 }
--- a/test/spec/features/context-pad/ContextPadSpec.js
+++ b/test/spec/features/context-pad/ContextPadSpec.js
@@ -16,6 +16,8 @@ import {
   classes as domClasses
 } from 'min-dom';
 
+import { getBBox } from 'lib/util/Elements';
+
 import contextPadModule from 'lib/features/context-pad';
 import selectionModule from 'lib/features/selection';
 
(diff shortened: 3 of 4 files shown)


assets/diagram-js.css
```diff
@@ -121,8 +121,13 @@
   stroke: var(--element-selected-outline-stroke-color);
 }
 
+.djs-connection.selected .djs-outline {
+  visibility: hidden;
```
nikku, line 125: Why not `display: none`? Do we rely on the outline for hover interactions?

lib/features/context-pad/ContextPad.js
```diff
@@ -513,6 +516,8 @@ ContextPad.prototype.isShown = function() {
  */
 ContextPad.prototype._getPosition = function(target) {
 
+  target = isConnection(target) ? getLastWaypoint(target) : target;
+
   var elements = isArray(target) ? target : [ target ];
```
nikku, line 521: Could we document what we're doing here? I think this is going to be confusing, especially mixed up with the `array vs. no array` logic.

assets/diagram-js.css
smbea: I was following our approach for not displaying the outlines: https://github.com/bpmn-io/diagram-js/pull/826/files/af76b072960f45fe0d4d3dc980ab44a6a0495fc1#diff-682322eea3bdf2a0256e2e37fb5a12ac5c9744f340488edcb053b4dbe8f57137R110-R112. But it works with `display` too

lib/features/context-pad/ContextPad.js
smbea: I added some phrases to the JSDocs: https://github.com/bpmn-io/diagram-js/pull/826/files#diff-6d674faf358204aeb4b308e49584a5b425dd2b59dcef70ebfb108f894b96db5dR513-R523. Is this what you had in mind?

assets/diagram-js.css
barmac: `visibility: hidden` means usually that we still want the element to affect the layout. However, since it's an SVG element within SVG, it doesn't really affect layout of anything. So it behaves exactly like `display: none` 🤷 .
https://developer.mozilla.org/en-US/docs/Web/CSS/visibility

lib/features/context-pad/ContextPad.js
```diff
@@ -507,12 +510,20 @@ ContextPad.prototype.isShown = function() {
 /**
  * Get contex pad position.
  *
+ * If target is a connection, the context pad will be placed according to the
+ * cnnection's last waypoint.
```
barmac, line 514:
```suggestion
 * connection's last waypoint.
 ```

File: lib/features/context-pad/ContextPad.js [hunk 4]
```diff
@@ -507,12 +510,20 @@ ContextPad.prototype.isShown = function() {
...
+ * If multiple targets, the context pad will be placed according to the bounding
+ * box containing all targets.
+ *
  * @param {ContextPadTarget} target
  *
  * @return {Rect}
  */
 ContextPad.prototype._getPosition = function(target) {
 
+  target = isConnection(target) ? getLastWaypoint(target) : target;
```
barmac, line 525: What if I pass a single-item array with connection?

barmac: So `getPad([ connection ]);`

smbea: We check for that here: https://github.com/bpmn-io/diagram-js/blob/0cfd9c77ca823f6eac4d549b95e1902817536893/lib/features/context-pad/ContextPad.js#L111

barmac: OK but that code is not run when I open context pad programmatically. I know that we are facing legacy logic here. Notice how we unwrap the target from array to wrap it again in `_getPosition`.

barmac: It looks like we don't handle programmatic opening of a single item array anyway, so perhaps we could refactor this context pad in a separate PR (some day maybe).

smbea: True, programatically it doesn't really make sense

assets/diagram-js.css
nikku: Affect layout and interaction (i.e. for mouse over). The user interaction may be more important here, for (invisible) click interaction boxes.

smbea: looks like something to be cleaned up some time 🙃

lib/features/context-pad/ContextPad.js
nikku: I think that makes stuff a lot clearer, thanks!

assets/diagram-js.css
barmac: TBH the whole discussion we had here suggests `display: none` is a better choice. That wouldn't make me think while `visibility` does.

smbea: Ok then let's go for the most understandable solution. I updated to use `display`

barmac (approved):

nikku:
Works! :)

PR #829, opened 2023-11-21
fix(popup): reposition if under axis

smbea:
Related to https://github.com/camunda/improved-canvas/pull/9

Currently, we don't consider that the popup menu can open upwards. I considered doing this in the actual extension but it leads to flaky positioning. And opening downwards is a bpmn-js concern, not diagram-js. So it's fair to also consider this case imo

Before:

https://github.com/bpmn-io/diagram-js/assets/25825387/9b632855-2cd9-4f1b-bd80-2415f7e1b92e

After:

https://github.com/bpmn-io/diagram-js/assets/25825387/6b01cdac-545a-4fc5-862d-0cf552530932

### DIFF
--- a/lib/features/popup-menu/PopupMenu.js
+++ b/lib/features/popup-menu/PopupMenu.js
@@ -341,6 +341,11 @@ PopupMenu.prototype._ensureVisible = function(container, position) {
     top = position.y - containerBounds.height;
   }
 
+  // underAxis
+  if (position.y < documentBounds.top) {
+    top = position.y + containerBounds.height;
+  }
+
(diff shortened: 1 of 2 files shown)


smbea:
In the weekly, I mentioned opening to bottom of context pad if there is no space on top. While I like that idea, it it's best to have a unified behaviour. We could look into not covering the context pad as a core improvement later

barmac:
I will review this today. Sorry for delay.

barmac (approved):
Let's merge it. It solves the issue and it's tested.

PR #833, opened 2023-11-30
Update name and link for object-diagram-js

timKraeuter:
Update name and link for object-diagram-js (previously object diagram modeler).


timKraeuter:
Huge thanks to you and the team at bpmn.io!

I am mostly just adopting tools from the rich bpmn.io ecosystem.

Issue #834, opened 2023-12-04
Allow non-searchable entries in popup menu


smbea:
### What should we do?

Support non-searchable entries in popup menu. For example, by providing `"searchable: false` in the entry object

### Why should we do it?

Allows groups with repeated entries like "suggested" or "common"

PR #835, opened 2023-12-04
feat(popup): allow non-searchable entries

smbea:
Closs #834

### DIFF
--- a/lib/features/popup-menu/PopupMenuComponent.js
+++ b/lib/features/popup-menu/PopupMenuComponent.js
@@ -77,6 +77,10 @@ export default function PopupMenuComponent(props) {
         return (entry.rank || 0) >= 0;
       }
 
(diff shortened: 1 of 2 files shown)


lib/features/popup-menu/PopupMenuComponent.js
```diff
@@ -77,6 +77,10 @@ export default function PopupMenuComponent(props) {
         return (entry.rank || 0) >= 0;
       }
 
+      if (entry.searchable === false) {
+        return 0;
```
nikku, line 81:
```suggestion
        return false;
```

Let's keep this interface uniform. We always return a `boolean`.

smbea: You're right, got momentarily confused with the `0` from above. Updated

nikku (approved):

PR #857, opened 2024-02-08
docs: add documentation generation

### LINKED ISSUE #78: Provide Documentation
Adding a (self) hosted solution with up to date documentation.

[**ReadMe.io**](https://readme.io/) could be a great option since it's a well built, easy to use system and free for open source.

---

Helpful resources, despite the lack of documentation:

* [diagram-js example](https://github.com/bpmn-io/diagram-js/tree/master/example)
* [bpmn-js architecture walkthrough](https://bpmn.io/toolkit/bpmn-js/walkthrough/), covering many facets of diagram-js and how it builds the foundation of our tools
* [source code](https://github.com/bpmn-io/diagram-js/tree/master/lib), including comprehensive 
code docs

joeskeen:
This PR takes all the awesome code documentation available in JSDoc/TSDoc comments throughout the code base and generates a documentation static site from it using the TypeDoc command-line tool.

A few helpful commands have been added to `package.json`:

* `generate-docs`: Runs TypeDoc and outputs the docs in the `docs` folder (ignored from source control)
* `publish-docs`: Generates the docs, goes into the docs folder, and commits/pushes the docs up to the `gh-pages` branch of the repository (by default, goes to `bpmn-io/diagram-js`, but you can pass another repository in to publish them elsewhere). To test, I ran this with `joeskeen/diagram-js` and it generated the GitHub Pages site at https://joeskeen.github.io/diagram-js

Changes to the actual content of the documentation is out of scope for this PR, just trying to publish all existing API documentation.

Closes #78

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules/
 dist/
+docs/
 coverage/
 lib/**/*.d.ts
 .idea/
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,6 +45,7 @@
         "puppeteer": "^22.0.0",
         "sinon": "^17.0.0",
         "sinon-chai": "^3.7.0",
+        "typedoc": "^0.25.7",
         "typescript": "^5.0.0",
         "webpack": "^5.74.0"
       },
@@ -1240,6 +1241,12 @@
       "engines": {
         "node": ">=8"
       }
+    },
+    "node_modules/ansi-sequence-parser": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/ansi-sequence-parser/-/ansi-sequence-parser-1.1.1.tgz",
+      "integrity": "sha512-vJXt3yiaUL4UU546s3rPXlsry/RnM730G1+HkpKE012AN0sx1eOrxSu95oKDIonskeLTijMgqWZ3uDEe3NFvyg==",
+      "dev": true
     },
     "node_modules/ansi-styles": {
       "version": "3.2.1",
@@ -4849,6 +4856,12 @@
         "node": ">=6"
       }
     },
+    "node_modules/jsonc-parser": {
+      "version": "3.2.1",
+      "resolved": "https://registry.npmjs.org/jsonc-parser/-/jsonc-parser-3.2.1.tgz",
+      "integrity": "sha512-AilxAyFOAcK5wA1+LeaySVBrHsGQvUFCDWXKpZjzaL0PqW+xfBOttn8GNtWKFWqneyMZj41MWF9Kl6iPWLwgOA==",
+      "dev": true
+    },
     "node_modules/jsonfile": {
       "version": "4.0.0",
       "resolved": "https://registry.npmjs.org/jsonfile/-/jsonfile-4.0.0.tgz",
@@ -5249,6 +5262,12 @@
         "node": ">=10"
       }
     },
+    "node_modules/lunr": {
+      "version": "2.3.9",
+      "resolved": "https://registry.npmjs.org/lunr/-/lunr-2.3.9.tgz",
+      "integrity": "sha512-zTU3DaZaF3Rt9rhN3uBMGQD3dD2/vFQqnvZCDv4dl5iOzq2IZQqTxu90r4E5J+nP70J3ilqVCrbho2eWaeW8Ow==",
+      "dev": true
+    },
     "node_modules/lz-string": {
       "version": "1.5.0",
       "resolved": "https://registry.npmjs.org/lz-string/-/lz-string-1.5.0.tgz",
@@ -5291,6 +5310,18 @@
       },
       "funding": {
         "url": "https://github.com/sponsors/sindresorhus"
+      }
+    },
+    "node_modules/marked": {
+      "version": "4.3.0",
+      "resolved": "https://registry.npmjs.org/marked/-/marked-4.3.0.tgz",
+      "integrity": "sha512-PRsaiG84bK+AMvxziE/lCFss8juXjNaWzVbN5tXAm4XjeaS9NAHhop+PjQxz2A9h8Q4M/xGmzP8vqNwy6JeK0A==",
+      "dev": true,
+      "bin": {
+        "marked": "bin/marked.js"
+      },
+      "engines": {
+        "node": ">= 12"
       }
     },
     "node_modules/media-typer": {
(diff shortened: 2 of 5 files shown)


nikku:
Your website rocks, and so do you. Thank you so much! 

We'll have a closer look beginning of next week.

nikku:
Some things I found having a shallow look:

* Would be great to link this back to source code somehow
* Classes are exposed as `default`, cf. [`CommandStack`](https://joe.skeen.rocks/diagram-js/classes/command_CommandStack.default.html) instead of their names.
* Some external dependencies are not documented; this will bite us in downstream libraries (i.e. bpmn-js) - https://joe.skeen.rocks/diagram-js/types/Diagram.ModuleDeclaration.html.
* Links don't seem to resolve consistently (https://joe.skeen.rocks/diagram-js/classes/core_ElementFactory.default.html#create.create-1)
* Really nice how well our existing documentation renders :tada:

joeskeen:
> Classes are exposed as default, cf. [CommandStack](https://joe.skeen.rocks/diagram-js/classes/command_CommandStack.default.html) instead of their names.
@nikku I have resolved this with #858

joeskeen:
> Would be great to link this back to source code somehow

@nikku I did look into this, and I'm finding that it's not possible currently with the way things work. From what I understand, you use your library `bio-dts` to generate the .d.ts files for each of the JS source files. But since the docs created via TypeDoc only work with TypeScript files, and since the .d.ts files are not checked in, there is no way to refer to the source. The TypeDoc tool does create references, but they are to the .d.ts files which don't resolve as links.

image

It doesn't appear that `bio-dts` creates source maps for the symbols in the generated files, but if it could, we may be able to map those links back to the original source.

Another alternative would be to use `jsdoc` instead of `typedoc`. I tried this out and found that the output was really not great compared to what we are getting through `typedoc`. My guess is that your `bio-dts` tool does a better job understanding the JS source than `jsdoc` does, so `typedoc` has a better input as a result. Thoughts?

joeskeen:
> Some external dependencies are not documented; this will bite us in downstream libraries (i.e. bpmn-js) - https://joe.skeen.rocks/diagram-js/types/Diagram.ModuleDeclaration.html.

@nikku I believe this is because the `.d.ts` files generated do not have any doc comments on the type aliases from the `didi` package (or others). I would expect the only way to fix this would be to alter the `bio-dts` process to add doc comments to those encountered `@typedefs`. See https://stackoverflow.com/a/62243905

joeskeen:
> Links don't seem to resolve consistently (https://joe.skeen.rocks/diagram-js/classes/core_ElementFactory.default.html#create.create-1)

I'm not sure what you mean. In the example you provided above, it seems that the link resolution works as follows:

* https://joe.skeen.rocks/diagram-js/classes/core_ElementFactory.default.html#create - the `create` method
* https://joe.skeen.rocks/diagram-js/classes/core_ElementFactory.default.html#create.create-1 - the first overload of the `create` method (taking the type `"label"`)
* https://joe.skeen.rocks/diagram-js/classes/core_ElementFactory.default.html#create.create-2 - the second overload of the `create` method (taking the type `"connection"`)
* https://joe.skeen.rocks/diagram-js/classes/core_ElementFactory.default.html#create.create-3 - the third overload of the `create` method (taking the type `"shape"`)
* https://joe.skeen.rocks/diagram-js/classes/core_ElementFactory.default.html#create.create-4 - the fourth overload of the `create` method (taking the type `"root"`)

nikku:
Thanks for your follow-ups on this topic. Find some comments attached:

> My guess is that your bio-dts tool does a better job understanding the JS source than jsdoc does, so typedoc has a better input as a result. Thoughts?

This is my thinking, too. We could do some magic at some point to find back-references to the source code from type definitions. Not impossible, but also not a priority today. Types are strictly validated and non-ambiguous, hence it makes sense to use them as a baseline.

nikku:
Regarding generation of external API references:

> [ref](https://github.com/bpmn-io/diagram-js/pull/857#issuecomment-1934657917): As this symbol is not exported, it does not get included in the documentation. Having bio-dts export that type would help.

I'm not sure it would help, as just me getting to the `didi` source code repository does not give me the assistance that I need. Why I'm especially picky about this is because many folks using our tools (and generating documentation) will face the same issue: You cannot document what is built on top of diagram-js (i.e. [bpmn-js](https://github.com/bpmn-io/bpmn-js)) without an easy way to source diagram-js documentation.

nikku:
> > Links don't seem to resolve consistently (https://joe.skeen.rocks/diagram-js/classes/core_ElementFactory.default.html#create.create-1)
> 
> I'm not sure what you mean. In the example you provided above, it seems that the link resolution works as follows [...]

Point taken, I think it works as expected. 

Unrelated to this the [file](https://joe.skeen.rocks/diagram-js/classes/core_ElementFactory.ElementFactory.html) is another showcase for how we need to export _all_ types declared (and used), or we'll not get proper documentation for them (https://github.com/bpmn-io/diagram-js/pull/857#issuecomment-1934657917, https://github.com/bpmn-io/diagram-js/pull/857#issuecomment-1945558707):

image

nikku:
To summarize what I learned through your investigations:

* exports (`default`) works as expected. We can fine tune the type definitions (https://github.com/bpmn-io/diagram-js/pull/858)
* we don't offer source maps which makes it impossible to map types back to JavaScript sources
* we'd need to export all types (also utility types) or we'll not get the documentation generation we need
* linking properly to external documentation is a topic of further exploration

nikku:
I did some more investigation on `typedoc` documentation generation yesterday:

> we don't offer [source maps for type declarations](https://stackoverflow.com/questions/48711065/go-to-the-typescript-source-file-instead-of-the-type-definition-file-in-vs-code) which makes it impossible to map types back to JavaScript sources

This is something we _could_ accomplish, as the bio-dts foundations support it out of the box.

> we'd need to https://github.com/bpmn-io/diagram-js/pull/857#issuecomment-1934657917 (also utility types) or we'll not get the documentation generation we need

It is actually [worse than this](https://github.com/TypeStrong/typedoc/issues/2488). In fact we'd need to export _all transitive_ types, which is completely unreasonable for larger libraries. Typescript itself, when generating types, does not export transitive types.

Typescript offers a clear contract: Everything that is exposed as public API, and hintable in a code editor shall also be documented. The assumptions that `typedoc` comes with don't work for us, the plug-ins provided don't seem to work to generate better types, but cause endless duplication in the documentation.

Needs to be further assessed (probably simply build our own docs generator).

nikku:
Declaration maps are possible, in theory. In practice it takes some work for sources to be properly mapped [through our pipeline](https://github.com/nikku/bio-dts/pull/3).

nikku:
@joeskeen Thanks again for your work on this topic. We'll close this PR for now as it is not ready to be incorporated; we may use it as a future reference once we tackle documentation generation ourselves, probably by writing our own tool.

What we did in the mean time is to invest into advanced typing, cf. https://github.com/bpmn-io/bpmn-js/pull/2153, inside your editor :star_struck:.

PR #858, opened 2024-02-08
refactor: separate export default from function declaration

joeskeen:
## The issue

Currently all classes (I know they are not technically ES6 classes, but I'm sure you know what I mean) are the default export of the package, which is fine, but since the declaration reads `export default function ClassName(){}`, when the typings are generated and documentation from those typings, all the class names become named `default`. See this from https://joe.skeen.rocks/diagram-js/classes/core_ElementFactory.default.html:

image

This leads to a potentially confusing experience for those consuming typings or docs from the library.

## Proposed solution

This PR exports each class as its own export, but also sets the default export to that class. The only change should be an additional export with the name of the class attached. This enables documentation to work better, as the names will exist in the docs, as well as links between referenced types.

image

## How the code changes were made

In VSCode, I did a RegEx global search in `lib/**/*.js` for:

```
(\/\*\*\n(.*?\n)+? \*\/)\nexport default function ([^\(]+)\(
```

and replaced it with:

```
export default $3;\n\n$1\nexport function $3(
```

Related to #78 
Related to https://github.com/bpmn-io/diagram-js/pull/857

### DIFF
--- a/lib/Diagram.js
+++ b/lib/Diagram.js
@@ -2,6 +2,8 @@ import { Injector } from 'didi';
 
 import CoreModule from './core';
 
+export default Diagram;
+
 /**
  * @typedef {import('didi').InjectionContext} InjectionContext
  * @typedef {import('didi').LocalsMap} LocalsMap
@@ -99,7 +101,7 @@ function createInjector(options) {
  * @param {DiagramOptions} [options]
  * @param {Injector} [injector] An (optional) injector to bootstrap the diagram with.
  */
-export default function Diagram(options, injector) {
+export function Diagram(options, injector) {
 
   this._injector = injector = injector || createInjector(options);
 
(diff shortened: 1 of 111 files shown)


joeskeen:
After looking more at the docs, it seems we may also want to do this for interfaces and modules. Thoughts @nikku?

nikku:
@joeskeen thanks for your contribution. 

We'll not be able to merge it at this stage, but we're internally in discussion how to proceed both regarding this PR and the proposed [documentation generation](https://github.com/bpmn-io/diagram-js/issues/78).

nikku:
@joeskeen Thanks again for your work. I'm closing this PR as I don't see us progressing on it any time soon. There is a bunch of things that we need to clarify, and likely we'll simply build a custom, opinionated documentation generator for our (huge) code base.

PR #859, opened 2024-02-14
docs: Update Canvas.js

joeskeen:
I was trying to use `canvas.zoom('fit-viewport')` but wanted it centered. Looking at the code, it appears that the second parameter can be `true` in this case to have it automatically center (rather than providing a point). Updating the documentation to reflect this.

### DIFF
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -1318,7 +1318,9 @@ Canvas.prototype.scrollToElement = function(element, padding) {
  *
  * @param {number|'fit-viewport'} [newScale] The new zoom level, either a number,
  * i.e. 0.9, or `fit-viewport` to adjust the size to fit the current viewport.
- * @param {Point} [center] The reference point { x: ..., y: ...} to zoom to.
+ * @param {Point|true} [center] The reference point { x: ..., y: ...} to zoom to, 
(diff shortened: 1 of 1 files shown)


nikku:
Thanks for the contribution. I believe `center` is an optional argument, and if not provided zooming (`fit-viewport`) will be done focused on the center.

I wonder if it makes sense to have a [dedicated type definition](https://github.com/bpmn-io/bpmn-js/blob/develop/lib/features/modeling/ElementFactory.js#L103) for the `fit-viewport` case, as `center = false` yields won't center there, but _will_ center for `zoom(0.3)`.

joeskeen:
Interesting. If I understand your last comment, it is expected to center the diagram if the second parameter is not provided? When I tried 'fit-viewport' without a second parameter, my diagram was always ending up at the far left.

nikku:
> If I understand your last comment, it is expected to center the diagram if the second parameter is not provided?

Yes. At best we _always_ center. I don't see a reason why `fit-viewbox` should not center the diagram (but push it to the coordinate origin).

nikku:
As indicated I'd love to have a dedicated override for the 'fit-viewport' case. https://github.com/bpmn-io/diagram-js/pull/859#issuecomment-1945517219.

PR #862, opened 2024-02-24
Be able to type diagram and events

nikku:
This allows implementors to type the diagram (available services) and eventBus events (available events):

### Sublime Text

capture hj3ree_optimized

### VSCode

capture WrEJZW_optimized

### DIFF
--- a/lib/core/EventBus.js
+++ b/lib/core/EventBus.js
@@ -118,6 +118,8 @@ var slice = Array.prototype.slice;
  * var sum = eventBus.fire('sum', 1, 2);
  * console.log(sum); // 3
  * ```
+ *
+ * @template [EventMap=null]
  */
 export default function EventBus() {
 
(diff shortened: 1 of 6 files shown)


lib/core/EventBus.js
```diff
@@ -131,8 +133,9 @@ export default function EventBus() {
   this.on('diagram.destroy', 1, this._destroy, this);
 }
 
-
 /**
+ * @overlord
```
vsgoulart, line 137: @nikku  I'm not familiar with this keyword and couldn't find it anywhere? isn't the correct one `@overload`?

nikku: This keyword is the working version of TypeScript `@overload`, cf. [documentation](https://github.com/nikku/bio-dts?tab=readme-ov-file#features).

nikku: If you want to use `@overlord`, then you'd need to use `bio-dts` to generate the types. 

We do it in core bpmn.io as our sources remain `.js` and we still want to have clean types.

barmac:
I will have a look at this today.

barmac (approved):
This works great. I can't wait to see that in action in bpmn-js.

PR #865, opened 2024-03-11
Fix types

nikku:
Fixes what is uncovered by running `tsc --skipLibCheck false`.

----

Depends on https://github.com/bpmn-io/diagram-js-ui/pull/23 if we want to execute our type checks with `skipLibCheck=true`.

### DIFF
--- a/lib/features/outline/Outline.js
+++ b/lib/features/outline/Outline.js
@@ -23,6 +23,7 @@ var DEFAULT_PRIORITY = 1000;
 /**
  * @typedef {import('../../model/Types').Element} Element
  *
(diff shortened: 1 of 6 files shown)


barmac:
I will update the dependency and push it here.

barmac:
@nikku Let's include https://github.com/bpmn-io/diagram-js/pull/866/commits/da7dd65d3f4620f6fd1e4d566a3e839132e424c0 in this PR.

barmac:
I pushed it on top.

barmac (approved):

PR #869, opened 2024-03-12
chore: import explicit files

nikku:
This prepares us to migrate diagram-js to an ES module.

Related to https://github.com/bpmn-io/diagram-js/issues/863

### DIFF
--- a/lib/Diagram.js
+++ b/lib/Diagram.js
@@ -1,6 +1,6 @@
 import { Injector } from 'didi';
 
-import CoreModule from './core';
+import CoreModule from './core/index.js';
 
 /**
  * @typedef {import('didi').InjectionContext} InjectionContext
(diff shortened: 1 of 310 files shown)


nikku:
Some tests are failing. I'm investigating why.

nikku:
All fixed!

nikku:
Validated by https://github.com/bpmn-io/diagram-js/issues/863 that this also works with strict (ESM) import checking.

barmac:
@nikku is this a breaking change? I think it isn't since the bundlers (which were previously required) can handle ESM, correct?

barmac (approved):

barmac:
> @nikku is this a breaking change? I think it isn't since the bundlers (which were previously required) can handle ESM, correct?

Confirmed on my own that there's no breaking change. It works just fine with bpmn-js via `npx @bpmn-io/sr bpmn-io/bpmn-js -l bpmn-io/diagram-js#esm`

nikku:
This just changes internal references - externally facing things are not affected. Thanks for the review :star:

PR #873, opened 2024-03-19
chore: document some arguments as optional

makepanic:
Hi,

we're in the process of upgrading from our own typedefs to the official diagram-js ones and found some cases where arguments are theoretically optional but documented as required in the jsdoc.

This PR adjusts the cases we noticed our build complaining.
The changes are usually where the argument is consumed with a fallback value in the javascript code, e.g. `argument || {}`.

If you expect these arguments not to be optional in the generated types, I can also close the PR.

Thanks


nikku (approved):
All of these changes make sense. Thanks!

nikku:
Released with [v14.2.0](https://github.com/bpmn-io/diagram-js/releases/tag/v14.2.0)

PR #874, opened 2024-03-21
A11y popup menu

### LINKED ISSUE #871: Make popup menu WCGA 2.1 compliant
### Is your feature request related to a problem? Please describe.

Currently the popup menu is a keyboard trap that is insufficiently indicated to assistive technology (screen readers), it violates [`no-keyboard-trap`](https://www.w3.org/WAI/WCAG22/quickref/#no-keyboard-trap).

### Describe the solution you'd like

The popup editor is clearly indicated as a keyboard trap with `ESC` (cancel) or `ENTER` (to select action) being the two promoted navigation actions.

### Describe alternatives you've considered

None.

### Additional context

Found via https://github.com/camunda/web-modeler/issues/7929#issuecomment-2003338026.

nikku:
#### Firefox

capture 5fPyRh_optimized

#### Chrome

capture 5w5PjG_optimized

---

This PR improves the accessibility of the popup menu:

* Ensures all items are keyboard accessible
* Ensures that existing TAB and arrow key navigation continues to work
* Ensures that `SPACE` and `ENTER` can be used as expected to toggle features
* Ensure that outside click is no longer trapped
* Use browser native outline for keyboard focus state

---

Additional notes:

* I decided that tabbing through the whole component (all entries) shall be possible
* I decided against making this a partial select/fake drop down from aria perspective, as it does not work in scenarios without search

----

Can be tried out using the following command on your local machine:

```
npx @bpmn-io/sr bpmn-io/bpmn-js -l bpmn-io/diagram-js#a11y-popup-menu
```

----

Closes #871

### DIFF
--- a/lib/features/popup-menu/PopupMenuComponent.js
+++ b/lib/features/popup-menu/PopupMenuComponent.js
@@ -65,8 +65,6 @@ export default function PopupMenuComponent(props) {
     return originalEntries.length > 5;
   }, [ search, originalEntries ]);
 
-  const inputRef = useRef();
-
   const [ value, setValue ] = useState('');
 
   const filterEntries = useCallback((originalEntries, value) => {
@@ -117,28 +115,6 @@ export default function PopupMenuComponent(props) {
   useEffect(() => {
     updateEntries(filterEntries(originalEntries, value));
   }, [ value, originalEntries ]);
-
-  // register global <Escape> handler
-  useEffect(() => {
-    const handleKeyDown = event => {
-      if (event.key === 'Escape') {
-        event.preventDefault();
-
-        return onClose();
-      }
-    };
-
-    document.documentElement.addEventListener('keydown', handleKeyDown);
-
-    return () => {
-      document.documentElement.removeEventListener('keydown', handleKeyDown);
-    };
-  }, []);
-
-  // focus input on initial mount
-  useLayoutEffect(() => {
-    inputRef.current && inputRef.current.focus();
-  }, []);
 
   // handle keyboard seleciton
   const keyboardSelect = useCallback(direction => {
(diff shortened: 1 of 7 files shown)


nikku:
To be considered: Do not [browser focus existing element](https://github.com/bpmn-io/diagram-js/pull/874/commits/4a72c06a6ede8efaffdaf95b4542884d8e26b2c7), but only do so on tab navigation:

capture r7pQV7_optimized

nikku:
As per discussion discarded the automatic keyboard focus:

* Ensure that resonable item is intially focused

---

capture 4dT3A2_optimized

barmac:
It looks good on Chrome, but there are some problems on Safari:

https://github.com/bpmn-io/diagram-js/assets/28307541/7c9ebc06-f333-462e-9e6c-e48b76ceae91

* The outline is cut-off.
* I cannot focus headers buttons with keyboard.
* The menu is blurred when zoomed in o.O (out of scope).

barmac:
This is Safari 17.0

barmac:
On Firefox, the outline is also cut off, but at least I can focus the header buttons, and the menu is never blurred:

image

nikku:
So we have to decide if we invent a completely new outline, standardized across all browsers.

I'd keep that as a follow-up activity.

barmac:
I'm OK with outline cut-off but it would be great if the focus worked correctly. I will check in the dev tools what is happening there.

barmac:
So via TAB I cannot reach the header modules, and after I TAB out, it focusses the body.

lib/features/popup-menu/PopupMenuItem.js
```diff
@@ -33,10 +33,14 @@ export default function PopupMenuItem(props) {
...
       data-id=${ entry.id }
       title=${ entry.title || entry.label }
+      tabIndex="0"
       onClick=${ onAction }
+      onFocus=${ onMouseEnter }
+      onBlur=${ onMouseLeave }
       onMouseEnter=${ onMouseEnter }
       onMouseLeave=${ onMouseLeave }
       onDragStart=${ (event) => onAction(event, entry, 'dragstart') }
+      aria-role="button"
```
barmac, line 43: If the items serve as buttons now, why don't we make them actual buttons? Spans make sense if we want to keep the select-option-like behavior but we decided against this as I understand.

lib/features/popup-menu/PopupMenuComponent.js
```diff
@@ -296,26 +260,58 @@ function PopupMenuWrapper(props) {
...
-    popupRef.current && popupRef.current.focus();
+    const popupEl = popupRef.current;
+
+    if (!popupEl) {
+      return;
+    }
+
+    const inputEl = popupEl.querySelector('input');
+
+    (inputEl || popupEl).focus();
```
barmac, line 273: I'd rather implement this via `autofocus` attribute.

barmac:
OK it might be that I don't know my tool. I can tab to the header buttons via Option+TAB: https://stackoverflow.com/questions/1848390/safari-ignoring-tabindex

barmac (approved):
LGTM

PR #875, opened 2024-03-21
chore: add some additional type corrections

makepanic:
Hi, 

I found some additional type incompatibilities when migrating to the most recent diagram-js version.
Most of them seem to be non issues and just documenting incorrect types. Some of them were found due to intellij complaining that e.g. a switch uses a case which isn't possible according to the types.

I'm unsure if the `Canvas.spec.ts` `canvas.scroll` call previously actually tested something, because the x/y parameters weren't used anywhere in the `canvas.scroll` method.

The `setRootElement` return value is only there because the return type specifies it always returns something but in that case returned `undefined`

### DIFF
--- a/lib/core/Canvas.js
+++ b/lib/core/Canvas.js
@@ -75,6 +75,7 @@ import { createMatrix as createMatrix } from 'tiny-svg';
  * @typedef {import('../util/Types').Point} Point
  * @typedef {import('../util/Types').Rect} Rect
  * @typedef {import('../util/Types').RectTRBL} RectTRBL
+ * @typedef {import('../util/Types').ScrollDelta} ScrollDelta
  */
 
 function round(number, resolution) {
@@ -775,7 +776,7 @@ Canvas.prototype.removeRootElement = function(rootElement) {
 Canvas.prototype.setRootElement = function(rootElement) {
 
   if (rootElement === this._rootElement) {
-    return;
+    return rootElement;
   }
 
   let plane;
@@ -1210,7 +1211,7 @@ Canvas.prototype.viewbox = function(box) {
 /**
  * Gets or sets the scroll of the canvas.
  *
- * @param {Point} [delta] The scroll to be set.
+ * @param {ScrollDelta} [delta] The scroll to be set.
  *
  * @return {Point}
  */
--- a/lib/core/Canvas.spec.ts
+++ b/lib/core/Canvas.spec.ts
@@ -158,7 +158,7 @@ canvas.removeConnection(connectionLike);
 
 canvas.resized();
 
-canvas.scroll({ x: 100, y: 100 });
+canvas.scroll({ dx: 100, dy: 100 });
 
 canvas.scrollToElement('shape');
 
--- a/lib/features/editor-actions/EditorActions.js
+++ b/lib/features/editor-actions/EditorActions.js
@@ -242,9 +242,9 @@ EditorActions.prototype.unregister = function(action) {
 };
 
 /**
- * Returns the number of actions that are currently registered
+ * Returns the identifiers of all currently registered editor actions
  *
- * @return {number}
+ * @return {string[]}
  */
 EditorActions.prototype.getActions = function() {
   return Object.keys(this._actions);
--- a/lib/util/Types.ts
+++ b/lib/util/Types.ts
@@ -3,6 +3,11 @@ export type Point = {
   y: number;
 };
 
+export type ScrollDelta = {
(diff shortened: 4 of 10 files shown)


lib/core/Canvas.js
```diff
@@ -775,7 +776,7 @@ Canvas.prototype.removeRootElement = function(rootElement) {
 Canvas.prototype.setRootElement = function(rootElement) {
 
   if (rootElement === this._rootElement) {
-    return;
+    return rootElement;
```
nikku, line 779: Great find. Could you [add a test case](https://github.com/bpmn-io/diagram-js/blob/develop/test/spec/core/CanvasSpec.js#L584) for this?

lib/core/Canvas.spec.ts
```diff
@@ -158,7 +158,7 @@ canvas.removeConnection(connectionLike);
 
 canvas.resized();
 
-canvas.scroll({ x: 100, y: 100 });
```
nikku, line 161: Not sure how that passed the type check. We'd want to ensure that either `dx` _or_ `dy` are set.

lib/features/editor-actions/EditorActions.js
```diff
@@ -244,7 +244,7 @@ EditorActions.prototype.unregister = function(action) {
 /**
  * Returns the number of actions that are currently registered
```
nikku, line 245:
```suggestion
 * Returns the identifiers of all currently registered editor actions
```

lib/util/Types.ts
```diff
@@ -3,6 +3,11 @@ export type Point = {
   y: number;
 };
 
+export type ScrollDelta = {
+  dx: number;
```
nikku, line 7:
```suggestion
  dx?: number;
```

lib/util/Types.ts
```diff
@@ -3,6 +3,11 @@ export type Point = {
...
+  dy: number;
```
nikku, line 8:
```suggestion
  dy?: number;
```

nikku (changes requested):
Great work! Find some remarks attached.

lib/core/Canvas.js
makepanic: Added a test for this https://github.com/bpmn-io/diagram-js/pull/875/files#diff-11db5d0fde9080dce33b2892042ad3e3913973a2bc59a46d784d592041df3e07R600-R611

lib/core/Canvas.spec.ts
makepanic: It previously passed because delta was typed incorrectly as Point.
Dunno if checking at runtime would be needed as it seems to default to some acceptable values in case the use isn't setting dx/dy.

test/spec/core/CanvasSpec.js
```diff
@@ -597,6 +597,19 @@ describe('Canvas', function() {
       expect(rootElement).to.equal(rootElement);
     }));
 
+    it('should return the current root element when attempting to set it again', inject(function(canvas, elementRegistry) {
+
+      // given
+      var rootElement = canvas.setRootElement({ id: 'XXXX' });
+
+      // when
+      rootElement = canvas.setRootElement(rootElement);
```
nikku, line 606:
```suggestion
      var updatedRootElement = canvas.setRootElement(rootElement);
```

test/spec/core/CanvasSpec.js
```diff
@@ -597,6 +597,19 @@ describe('Canvas', function() {
...
+
+      // then
+      // is the current root element returned from setter?
+      expect(rootElement).to.equal(rootElement);
```
nikku, line 610:
```suggestion
      expect(updatedRootElement).to.equal(rootElement);
```

test/spec/core/CanvasSpec.js
```diff
@@ -597,6 +597,19 @@ describe('Canvas', function() {
       expect(rootElement).to.equal(rootElement);
     }));
 
+    it('should return the current root element when attempting to set it again', inject(function(canvas, elementRegistry) {
```
nikku, line 600:
```suggestion

    it('should return the current root element when attempting to set it again', inject(function(canvas, elementRegistry) {
```

nikku: Spacing (we use two lines between tests to promote readability).

nikku (approved):
Good work!

nikku:
(And thanks for another nice contribution :medal_sports:).

PR #876, opened 2024-03-22
fix(PopupMenu): show `emptyPlaceholder` when no entries exist

marstamm:
This allows us to show empty placeholder when no entries were returned (e.g. still fetching), not just when the search filters out all entries

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ All notable changes to [diagram-js](https://github.com/bpmn-io/diagram-js) are d
 
 _**Note:** Yet to be released changes appear here._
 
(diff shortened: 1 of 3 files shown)


nikku (commented):
If I understand this unlocks the following use-case:

* Regardless of whether search is active or not, if no entries exist then the `emptyPlaceholder` is shown.

Is that correct?

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ All notable changes to [diagram-js](https://github.com/bpmn-io/diagram-js) are d
 
 _**Note:** Yet to be released changes appear here._
```
nikku, line 8: The component is `popup-menu`, please adjust the commit accordingly.

nikku:
Fine to merge then with `popup-menu` as the touched upon component. I'll cut a release soon and will include this change.

marstamm:
Already adjusted the commit message

PR #878, opened 2024-04-04
fix(popup-menu): keep given top-left position with scale

marstamm:
This ensures the position given to the Popup menu when opening (x, y) are the real coordinates it is displayed at.

Currently, the scale origin is in the center of the container, which creates weird UI when you zoomed in to the canvas:
image

With the fix, the transform is applied to the corner instead:
image


nikku:
Is this related to https://github.com/camunda/improved-canvas/issues/37?

marstamm:
I did not have this issue on my radar. It should be the same issue.

I can't replicate the Screenshot 100%, but the issue is present in dmn-js and this fix also affects it:

image

vs.

image

nikku (approved):
Tested and confirmed this works as expected. Good job!

PR #882, opened 2024-04-12
fix(ToolManager): do not force-cancel dragging

### LINKED ISSUE #881: Editor throws error when enabling dragging while a tool is active
See comment, wrong issue linked initially

Steps to Reproduce:

- press and hold `c` in the editor
- the connect tool should now be enabled/disabled rapidly
- click on an element
- The error is shown in the console

Root Cause:
This behavior causes a mismatch between Dragging state Active state in the tool manager. When we try to start a connection through clicking, dragging is initiated. Because we cancel the event (https://github.com/bpmn-io/diagram-js/blob/0ec9a1391e9c1f7ace7ad317fe3943a97dabdacc/lib/features/tool-manager/ToolManager.js#L87-L88) while we are initiating the drag operation, it breaks when we access the context immediately after.

Proposed Fix:
It is not clear why we cancel dragging in this place. This should be handled either by `dragging` or a higher component, such as `global-connect`. This is already the case.
No tests break in diagram-js and bpmn-js by just removing this logic, and with my tests, everything still works as expected.

marstamm:
closes #881

Reproduction instructions can be found in the linked ticket

### DIFF
--- a/lib/features/tool-manager/ToolManager.js
+++ b/lib/features/tool-manager/ToolManager.js
@@ -7,7 +7,6 @@ import {
 } from 'min-dom';
 
 /**
- * @typedef {import('../dragging/Dragging').default} Dragging
  * @typedef {import('../../core/EventBus').default} EventBus
  *
  * @typedef {import('../../core/EventBus').Event} Event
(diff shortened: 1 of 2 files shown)


nikku:
Found https://github.com/bpmn-io/bpmn-js/issues/2140 in the process of testing this feature.

nikku (approved):
What really bugs me is the text selection that happens during dragging: 

capture 7wQvji_optimized

I could not break the implementation otherwise, so let's go ahead and get this integrated.

marstamm:
> What really bugs me is the text selection

That seems to happen with the current main as well. +1 to track it, as this is unexpected behavior

PR #884, opened 2024-04-22
fix(types):  Palatte entry actions

miyuesc:


nikku (changes requested):
Thanks for your contribution :heart:.

Let's add a test case to verify this alternative usage: https://github.com/bpmn-io/diagram-js/blob/8403413c7ed99aac8e41b2263f9f29d9bdca8ceb/lib/features/palette/PaletteProvider.spec.ts#L27. 

Ensures that we'll not accidentally unbuild it.

nikku (approved):

PR #909, opened 2024-05-29
Prevent duplicate marker ids in preview

marstamm:
This PR:

- changes the marker ID if it was part of the `djs-visual` instead of creating a new one.
- Adds a random string to the copied marker ID. This ensures we do not have overlaps if the same connection is present multiple times in a preview. While this should not happen, it did happen in the past so better be safe here.
- Ensures we copy connections only once in move preview

related to https://github.com/bpmn-io/bpmn-js/issues/2179

### DIFF
--- a/lib/features/move/MovePreview.js
+++ b/lib/features/move/MovePreview.js
@@ -4,7 +4,6 @@ import {
   filter,
   find,
   groupBy,
-  map,
   matchPattern,
   size
 } from 'min-dash';
@@ -64,11 +63,14 @@ export default function MovePreview(
   function getAllDraggedElements(shapes) {
     var allShapes = selfAndAllChildren(shapes, true);
 
-    var allConnections = map(allShapes, function(shape) {
-      return (shape.incoming || []).concat(shape.outgoing || []);
-    });
+    var allConnections = allShapes.flatMap(shape =>
+      (shape.incoming || []).concat(shape.outgoing || [])
+    );
 
-    return flatten(allShapes.concat(allConnections));
+    var allElements = allShapes.concat(allConnections);
+    var uniqueElements = [ ...new Set(allElements) ];
+
+    return uniqueElements;
   }
 
   /**
--- a/lib/features/preview-support/PreviewSupport.js
+++ b/lib/features/preview-support/PreviewSupport.js
@@ -10,6 +10,8 @@ import { query as domQuery } from 'min-dom';
 
 import { getVisual } from '../../util/GraphicsUtil';
 
+import Ids from '../../util/IdGenerator';
+
 /**
  * @typedef {import('../../core/Types').ElementLike} Element
  * @typedef {import('../../core/Types').ShapeLike} Shape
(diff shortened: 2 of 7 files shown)


lib/features/preview-support/PreviewSupport.js
```diff
@@ -175,35 +173,32 @@ PreviewSupport.prototype._cloneMarkers = function(gfx, className = 'djs-dragger'
...
-  var markerId = marker.id + '-' + className;
-
-  var clonedMarker = this._clonedMarkers[ markerId ];
-
-  parentGfx = parentGfx || this._canvas._svg;
 
-  if (!clonedMarker) {
-    clonedMarker = svgClone(marker);
+  // Add a random suffix to the marker ID in case the same marker is previewed multiple times
+  var clonedMarkerId = [ marker.id, className, 'clone', crypto.randomUUID() ].join('-');
```
marstamm, line 178: We do not need to persist the Ids, as the preview is ephemeral. I decided to go for `crypto.randomUUID()` instead of the `Ids` library to save on performance + bundle size.
It is supported by all modern browsers
image

nikku: How can the same marker be previewed multiple times? I'd rather not depend on too many modern API tools. You could (alternatively) simply generate _one_ static preview ID and add an auto-incrementing number to it.

lib/features/move/MovePreview.js
```diff
@@ -68,7 +68,10 @@ export default function MovePreview(
       return (shape.incoming || []).concat(shape.outgoing || []);
     });
 
-    return flatten(allShapes.concat(allConnections));
+    var allElements = flatten(allShapes.concat(allConnections));
```
nikku, line 71: We could do some flatMap magic above^^.

lib/features/preview-support/PreviewSupport.js
nikku: We already have a cheap internal ID generator in place, cf. [`IdGenerator`](https://github.com/bpmn-io/diagram-js/blob/d1caba8b9d4c95f3aa3d53ca58e6bd44ce1c3ccc/lib/util/IdGenerator.js#L11)

marstamm: > How can the same marker be previewed multiple times?

See [7e2bfe3](https://github.com/bpmn-io/diagram-js/pull/909/commits/7e2bfe3bb14dea07998102dcdcb9ba868fd12cb7). When you add the same connection multiple times, the marker will end up with the same copied ID if we do not add randomness.
While it was a bug in this case, I can also see some advanced use-cases where this might be useful (collaborative editing on the same Object?). Let's be robust in the core.

lib/features/preview-support/PreviewSupport.js
```diff
@@ -20,6 +22,8 @@ import { getVisual } from '../../util/GraphicsUtil';
  * @typedef {import('../../draw/Styles').default} Styles
  */
 
+const cloneIds = new Ids('clone');
```
nikku, line 25:
```suggestion
const cloneIds = new Ids('ps');

```

nikku: We could ensure that the ids are scoped to the component. Overlays for example uses `ov` as a prefix.

marstamm: adjusted and squashed 🚀

PR #910, opened 2024-05-29
deps: bump to tiny-svg@3.1.2 / min-dom@4.2.1

nikku:
Improves [`clear`](https://github.com/bpmn-io/tiny-svg/blob/main/lib/clear.js).


nikku:
As it turns out using `Element.replaceChildren`, a modern API, did not yield expected performance gains. In fact the implementation was slightly slower (both reproduced through full suite tests, and local spacetool tests). Reverted the change with follow-up utility library releases.

marstamm (approved):

PR #915, opened 2024-06-05
fix(context-pad): only update visibility once per frame

marstamm:
Significantly improves deselection performance

related to https://github.com/camunda/camunda-modeler/issues/4335
Recording 2024-06-05 at 15 06 13

### DIFF
--- a/lib/features/context-pad/ContextPad.js
+++ b/lib/features/context-pad/ContextPad.js
@@ -31,6 +31,7 @@ var MARKER_HIDDEN = 'djs-element-hidden';
  * @typedef {import('../../core/Canvas').default} Canvas
  * @typedef {import('../../core/ElementRegistry').default} ElementRegistry
  * @typedef {import('../../core/EventBus').default} EventBus
+ * @typedef {import('../../scheduler/Scheduler').default} Scheduler
  *
  * @typedef {import('./ContextPadProvider').default} ContextPadProvider
  * @typedef {import('./ContextPadProvider').ContextPadEntries} ContextPadEntries
@@ -56,12 +57,14 @@ var HOVER_DELAY = 300;
  * @param {Canvas} canvas
  * @param {ElementRegistry} elementRegistry
  * @param {EventBus} eventBus
+ * @param {Scheduler} scheduler
  */
-export default function ContextPad(canvas, elementRegistry, eventBus) {
+export default function ContextPad(canvas, elementRegistry, eventBus, scheduler) {
 
   this._canvas = canvas;
   this._elementRegistry = elementRegistry;
   this._eventBus = eventBus;
+  this._scheduler = scheduler;
 
   this._current = null;
 
@@ -71,7 +74,8 @@ export default function ContextPad(canvas, elementRegistry, eventBus) {
 ContextPad.$inject = [
   'canvas',
   'elementRegistry',
-  'eventBus'
+  'eventBus',
+  'scheduler'
 ];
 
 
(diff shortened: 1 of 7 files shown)


lib/features/context-pad/ContextPad.js
```diff
@@ -622,25 +630,39 @@ ContextPad.prototype._updatePosition = function() {
  * using the `djs-element-hidden` or `djs-label-hidden` markers.
  */
 ContextPad.prototype._updateVisibility = function() {
-  if (!this.isOpen()) {
+
+  // Only update pad once per frame
+  // cf. https://github.com/camunda/camunda-modeler/issues/4335
+  if (this._requestedFrame) {
```
nikku, line 636: Shall we wrap this into a "scheduling utility"? I forsee that we'll need this in other places, too.

marstamm: I think so. We will have to make it available through dependency injection, since we want to do cancel on destroy centrally as well, which requires the `eventBus`.

nikku: Alternative: 

* Expose it explicitly as a task to execute; no eventBus and complicated logic needed

```
function MyComponent() {

  // ensure that update is only reset 
  this._updateTask = createTask((some, dependency) => {
    
  });
}

// execute it at any time
myComponent._updateTask.schedule(some, dependency);
```

marstamm: If we delay execution, we will need to be able to cancel it. I want to ensure this is reusable.

When integration testing with bpmn-js, I already ran into issues if we do not properly clean up, as we might try to access DOM elements that have been removed since we scheduled the task. Since this will most likely be relevant to more instances, I would like to have this logic central.

Alternative to dependency injection: make it a utility class and every Component that needs it can create a local Scheduler, with optional cleanup

marstamm: cf. [1e74dff](https://github.com/bpmn-io/diagram-js/pull/915/commits/1e74dff27438d9afb40a97f27d1b3069aeaab275) as a proposed utility

lib/features/context-pad/ContextPad.js
```diff
@@ -622,25 +626,35 @@ ContextPad.prototype._updatePosition = function() {
...
 ContextPad.prototype._updateVisibility = function() {
-  if (!this.isOpen()) {
-    return;
-  }
 
-  var self = this;
+  const updateFn = () => {
 
-  var target = this._current.target;
+    this._requestedFrame = null;
```
barmac, line 632: It looks that it's not used anymore.

```suggestion
```

marstamm: You are right, I missed it when refactoring 👍

nikku:
I've added my cents via https://github.com/bpmn-io/diagram-js/pull/915/commits/e3d1ebc3bec72d157d5521d31792daba7dcfa359, specifically the following:

* Make it a generally re-usable utility.
* Allow integrations to sensibly handle return values and errors. Don't swallow errors!
* Test without asserting internals

----

On a related note you _could_ argue that the actual bug is a broken ("not performant") `Canvas#hasMarker` method. As a user you're not supposed to fiddle with the CSS of the element anyway, so markers (tested via `hasMarker`) could be cached, too.

marstamm:
I think it is both. `Canvas#hasMarker` should not query the DOM and the context pad should not calculate visibility multiple times in one frame.

I added the changes to `hasMarker` to decouple it from the DOM in [6cefa79](https://github.com/bpmn-io/diagram-js/pull/915/commits/6cefa79365033a2131ca55db1bf09a4d711115a4)

lib/features/context-pad/ContextPad.js
```diff
@@ -637,25 +641,30 @@ ContextPad.prototype._updatePosition = function() {
...
+    });
+
+    if (isHidden) {
+      self.hide();
+    } else {
+      self.show();
+    }
+  };
+
+  this._scheduler.schedule(updateFn, 'ContextPad#updateVisibility');
```
nikku, line 667:
```suggestion
  this._scheduler.schedule(updateFn, 'ContextPad#_updateVisibility');
```

nikku:
Great! Spun off marker check to https://github.com/camunda/camunda-modeler/issues/4335.

nikku:
Cleaned up history, split both feature (new scheduler) + fix (update once per frame).

PR #918, opened 2024-06-21
fix: assign correct role attribute in popup menu item

barmac:
Related https://github.com/camunda/camunda-modeler/issues/4383

### Proposed Changes

This removes incorrect `aria-role` attribute from the popup menu item, and adds `role=button` to the list item child (the popup menu content). Note that the `<li>` element has an implicit role of [listitem](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listitem_role) and a `role=button` element as a direct child of a list is not allowed.

### Checklist


lib/features/popup-menu/PopupMenuItem.js
```diff
@@ -40,10 +40,9 @@ export default function PopupMenuItem(props) {
       onMouseEnter=${ onMouseEnter }
       onMouseLeave=${ onMouseLeave }
       onDragStart=${ (event) => onAction(event, entry, 'dragstart') }
-      aria-role="button"
       draggable=${ true }
     >
-      <div class="djs-popup-entry-content">
+      <div class="djs-popup-entry-content" role="button">
```
nikku, line 45: Any reason why you attach it to a non-interactive element? I'd rather attach it to the element that has click events registered.

barmac: > Note that the `<li>` element has an implicit role of [listitem](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listitem_role) and a role=button element as a direct child of a list is not allowed.

My first try was what you suggested, but then the tests failed due to unexpected "button" as a list child (should be listitem).

barmac: As an alternative, we could get rid of the role completely, as it does not seem to provide any value.

barmac: We could also rethink the popup menu semantics, and mark it as [`combobox`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role) if the search input is available, or as `menu` for the context menu. But it's another topic, much broader than what we see here.

barmac:
I removed the `role` attribute to keep the changes as minimal as possible. `aria-role` was already incorrect, so there are no changes in the a11y tree with this PR.

nikku (approved):

lib/features/popup-menu/PopupMenuItem.js
nikku: Makes sense, thanks for elaborating.

PR #922, opened 2024-07-04
921 firefox scheduler test fix

nikku:
### Checklist

* [ ] **Brief textual description** of the changes present

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules/
 dist/
+tmp/
 coverage/
 lib/**/*.d.ts
 .idea/
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -2,6 +2,8 @@
 
 var path = require('path');
 
+var fs = require('fs');
+
 var coverage = process.env.COVERAGE;
 
 // configures browsers to run test against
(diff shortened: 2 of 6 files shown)


nikku:
This can be reproduced somewhat locally using the [Firefox headless runner](https://github.com/bpmn-io/diagram-js/pull/922/commits/a4ccc1f2075413b1c9c20e7367f52cb146c8a561), but only when running the _whole_ test suite. Executing only the scheduler specs passes. 

Current assumption is that we don't properly cleanup the scheduler, or that Firefox swallows the [`requestAnimationFrame` calls](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame) (likely).

nikku:
Executing test on Firefox takes whooping 12 minutes:

capture J7ZIYn_optimized

barmac (approved):

barmac:
Hmm maybe we should not rush it. This causes some tests to fail on my machine (MacOS, Chrome Headless 129).

barmac:
I'll check on fresh instal.

barmac:
OK so I guess I am now suffering from the same problem as Firefox before the changes. On main, I can `npm run all` all green, but with this PR, the tests start to timeout :/

barmac:
And the tests are very old, rather unrelated to scheduler.

PR #948, opened 2024-11-11
fix(keyboard): bind-method typing still requires EventTarget

torge-hmn:
### Proposed Changes

In a previous update the ability to bind the keyboard to a specific node has been removed. The current JS-doc resulted in a generated type definition still requiring an `EventTarget` as parameter for the `bind` method even though supplying this parameter leads to an error and has no effect.

Prior (generated `Keyboard.d.ts`):

```
  /**
   * Bind keyboard events to the given DOM node.
   *
   * @param node
   */
  bind(node: EventTarget): void;
```

After:

```
  /**
   * Bind keyboard events to the given DOM node.
   *
   * @param node
   */
  bind(node: undefined): void;
```

### Checklist

* [x] **Visual demo** attached: See comparison above
* [x] **Steps to try out** present, i.e. [using the `@bpmn-io/sr` tool](https://github.com/bpmn-io/sr): Just use type generation
* [x] Related issue linked: I've not filed an issue yet

---

Related to https://github.com/bpmn-io/diagram-js/pull/662

### DIFF
--- a/lib/features/keyboard/Keyboard.js
+++ b/lib/features/keyboard/Keyboard.js
@@ -126,7 +126,7 @@ Keyboard.prototype._isEventIgnored = function(event) {
  * Bind keyboard events to the given DOM node.
  *
  * @overlord
- * @deprecated
+ * @deprecated No longer in use since version 15.0.0.
  *
  * @param {EventTarget} node
  */
--- a/lib/features/keyboard/Keyboard.js
+++ b/lib/features/keyboard/Keyboard.js
@@ -49,7 +49,7 @@ var compatMessage = 'Keyboard binding is now implicit; explicit binding to an el
  * `keyboard.bind=true|false` configuration option.
  *
  * @param {Object} config
- * @param {EventTarget} [config.bindTo]
+ * @param {boolean} [config.bind]
  * @param {EventBus} eventBus
  */
 export default function Keyboard(config, eventBus) {
@@ -125,10 +125,17 @@ Keyboard.prototype._isEventIgnored = function(event) {
 /**
  * Bind keyboard events to the given DOM node.
  *
+ * @overlord
+ * @deprecated
+ *
  * @param {EventTarget} node
+ */
+/**
+ * Bind keyboard events to the canvas node.
  */
 Keyboard.prototype.bind = function(node) {
 
+  // legacy <node> argument provided
   if (node) {
     console.error('unsupported argument <node>', new Error(compatMessage));
   }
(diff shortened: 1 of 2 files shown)


torge-hmn:
I would prefer removing the parameter from the generated typing entirely if it is possible. I don't know whether the type-generation tooling has a handling for this use case, as `any`-typing was used in a basic attempt I tried.

nikku:
CC @jarekdanielak I believe a follow-up is needed (this PR, to be checked) to properly move fix our type definitions, too.

nikku:
Thanks @torge-hmn for the PR :star: . We'll have a look.

lib/features/keyboard/Keyboard.js
```diff
@@ -125,7 +125,7 @@ Keyboard.prototype._isEventIgnored = function(event) {
 /**
  * Bind keyboard events to the given DOM node.
  *
- * @param {EventTarget} node
+ * @param {undefined} node
  */
 Keyboard.prototype.bind = function(node) {
```
nikku, line 130: We should rather fix the method signature, not require any argument, and check whether an argument is provided via `arguments[0]`. This change looks odd.

```suggestion
Keyboard.prototype.bind = function() {
```

nikku:
@jarekdanielak Ready for your review!

lib/features/keyboard/Keyboard.js
torge-hmn: Much better solution, thank you! 👍 (I wasn't sure if I could change the signature tbh)

lib/features/keyboard/Keyboard.js
```diff
@@ -125,10 +125,17 @@ Keyboard.prototype._isEventIgnored = function(event) {
 /**
  * Bind keyboard events to the given DOM node.
  *
+ * @overlord
+ * @deprecated
```
jarekdanielak, line 129:
```suggestion
 * @deprecated Since version 15.0.0, do not specify the `node` parameter anymore.
```

Perhaps we can have more informative deprecation message.

nikku: capture HEGYmm_optimized

jarekdanielak (approved):

nikku:
Thanks @torge-hmn for the report + initial fix.

torge-hmn:
> Thanks @torge-hmn for the report + initial fix.

Thanks for fixing! 👍

nikku:
Published via [diagram-js@15.2.2](https://github.com/bpmn-io/diagram-js/releases/tag/v15.2.2).

PR #953, opened 2024-11-21
fix(search): prioritize start of word and exact match more

nikku:
Test via `npx @bpmn-io/sr bpmn-io/bpmn-js -l bpmn-io/diagram-js#search-start`

### Proposed Changes

This prioritizes start of term and exact matches more aggressively.

_This yields more substantial results in large popup (append) menus:_

capture SO2PNd_optimized

_This yields better search results when querying larger data pools, where exact matches would otherwise de-prioritized:_

capture 5sTQ3Y_optimized

Related to https://github.com/camunda/camunda-modeler/issues/4122, https://github.com/camunda/camunda-modeler/issues/3439.

### Checklist

### DIFF
--- a/lib/features/search/search.js
+++ b/lib/features/search/search.js
@@ -181,10 +181,16 @@ function scoreToken(token) {
   }
 
   return token.start
-    ? 1.37
-    : token.wordStart
-      ? 1.13
-      : 1;
+    ? (
+      token.end
+        ? 131.9
+        : 7.87
+    )
+    : (
+      token.wordStart
+        ? 2.19
+        : 1
+    );
 }
 
 /**
@@ -223,7 +229,12 @@ function matchString(string, words) {
   const tokens = [];
   const matchedWords = {};
 
-  const regexpString = words.map(escapeRegexp).flatMap(str => [ '(?<wordStart>\\b' + str + ')', str ]).join('|');
+  const wordsEscaped = words.map(escapeRegexp);
+
+  const regexpString = [
+    `(?<all>${wordsEscaped.join('\\s+')})`,
+    ...wordsEscaped
+  ].join('|');
 
   const regexp = new RegExp(regexpString, 'ig');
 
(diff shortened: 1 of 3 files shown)


barmac:
I added the `sr` script to the top comment, and am looking into this right now.

lib/features/search/search.js
```diff
@@ -175,10 +175,16 @@ function scoreToken(token) {
...
-      : 1;
+    ? (
+      token.end
+        ? 131.9
+        : 7.87
+    )
+    : (
+      token.wordStart
+        ? 2.19
+        : 1
```
barmac, line 186: What magic is this? :) Why such values?

barmac:
While searching for the start of the word works better with this PR, the exact match with spaces is moved down:

This PR:

image

Current:

image

barmac:
I can reproduce https://github.com/camunda/camunda-modeler/pull/4711#issuecomment-2491044304 in this PR too.

lib/features/search/search.js
nikku: Standard magic values, call them temperature :hot_face:

barmac: Gives me AI engineer vibes :D

test/spec/features/search/searchSpec.js
```diff
@@ -253,6 +253,32 @@ describe('features/search', function() {
   }));
 
 
+  it('should prioritize longest match', inject(function(search) {
```
barmac, line 256: Good one!

nikku:
@barmac Thanks for your tests! I think we get to something reasonable, one test at a time. Check my latest updates on this branch out or the following capture that shows this is fixed: We now score tokens by _length_, too, to prioritize longest matches. Also we de-prioritize non-matches:

capture i7bkrt_optimized

barmac (approved):

lib/features/search/search.js
nikku: You gotta tune it some times. Though we have test coverage, which is good. :wink:

PR #957, opened 2025-01-09
FIx undo/redo for collapsed subprocesses

jarekdanielak:
### Proposed Changes

 

Related to https://github.com/bpmn-io/bpmn-js/issues/2269

Add some null checks to prevent crashes when undo/redo adds or removes a collapsed subprocess.

### Checklist

### DIFF
--- a/lib/core/GraphicsFactory.js
+++ b/lib/core/GraphicsFactory.js
@@ -192,6 +192,11 @@ GraphicsFactory.prototype.updateContainments = function(elements) {
     forEach(children.slice().reverse(), function(child) {
       var childGfx = elementRegistry.getGraphics(child);
 
+      // shape was removed
+      if (!childGfx) {
+        return;
+      }
+
       prependTo(childGfx.parentNode, childrenGfx);
     });
   });
--- a/lib/features/modeling/cmd/MoveConnectionHandler.js
+++ b/lib/features/modeling/cmd/MoveConnectionHandler.js
@@ -58,7 +58,7 @@ MoveConnectionHandler.prototype.revert = function(context) {
       delta = context.delta;
 
   // remove from newParent
-  collectionRemove(newParent.children, connection);
+  newParent && collectionRemove(newParent.children, connection);
 
   // restore previous location in old parent
   collectionAdd(oldParent.children, connection, oldParentIndex);


barmac:
Do you have a branch with an integration test for bpmn-js? Let's verify that this PR fixes the issue when linked with bpmn-js.

barmac (changes requested):
Please add tests (in diagram-js too) which reproduce the issue and allow us to verify the fix.

jarekdanielak:
I extended the integration tests in bpmn-js with one that covers this bug. Failing with current version, passes with the change: https://github.com/bpmn-io/bpmn-js/commit/5d8240d1afe801ee5e144cdca2b3680cdc94f2dc

lib/core/GraphicsFactory.js
```diff
@@ -192,6 +192,11 @@ GraphicsFactory.prototype.updateContainments = function(elements) {
     forEach(children.slice().reverse(), function(child) {
       var childGfx = elementRegistry.getGraphics(child);
```
nikku, line 193: I struggle to understand in which situation this can become an issue, could you elaborate? And would you be able to model this as a test case in diagram-js?

From what I understand: 

* `GraphicsFactory#updateContainment` is being called after any change operation happened ([ref](https://github.com/bpmn-io/diagram-js/blob/371877b77a548525d58ac12b56143eec3eaf17b4/lib/features/change-support/ChangeSupport.js#L57))
* At this stage, looking at the model, parents for updated children are collected
* Then the children inside these parents are re-ordered
* _BUT_ some of the children are actually _NOT_ part of the render tree anymore?

The _BUT_ is the big issue for me. How can children exist, but not actually be rendered? Where don't we properly update the model _before_ we render?

nikku (changes requested):
@jarekdanielak Let's [figure out if we fixed the root cause](https://github.com/bpmn-io/diagram-js/pull/957#discussion_r1910347319) of the issue here, happy to help with a debug session / code review.

jarekdanielak:
I spent some time debugging: 

- `GraphicsFactory#updateContainment` crashes because an element inside of a subprocess that is being deleted by undo still has a parent assigned (the subprocess).
- Element's parent is correctly unset by `Canvas#_removeElement`, but is later brought back by `MoveShapeHandler#revert`. 
- `MoveShapeHandler#revert` is fired because `elements.move` is fired when creating a collapsed subprocess.

At this point I need some help. I don't know why we call `elements.move` when creating a new collapsed subprocess. It doesn't fire when creating other types of elements. 

@nikku could you step in?

barmac:
@jarekdanielak I'm happy to support you with this. We can have a debugging session together to root cause the problem.

nikku:
Thanks @barmac, will remove this from my TODO list. If you folks cannot figure it out, please ping me again.

jarekdanielak:
Debugging session conclusions so far:

- When collapsed subprocess is created, we [first make sure its children are visible](https://github.com/bpmn-io/bpmn-js/blob/0a6e428ff3f7586d9dd279c32bd85c208d669359/lib/features/modeling/behavior/SubProcessPlaneBehavior.js#L118).
- Then children are [moved and positioned in the newly created plane](https://github.com/bpmn-io/bpmn-js/blob/0a6e428ff3f7586d9dd279c32bd85c208d669359/lib/features/modeling/behavior/SubProcessPlaneBehavior.js#L120).
- When user triggers undo, `shape.create` action for each element is reverted therefor they are removed.
- Then, `shape.move` action is reverted for each child. The child element's parent was set to `null` by `Canvas#_removeElement`, but move action revert handler sets it back to subprocess.

barmac:
The SubProcessPlaneBehavior is triggered on each `shape.create`. However, we could trigger this only once for `elements.create` command. This means that we would move the elements at the end of the action. We are trying this out.

barmac:
There's more in this issue than we expected. @jarekdanielak will push our solution sketch with the still not working text annotation part. The diagram-js PR can be closed as we established that the root cause is in bpmn-js.

jarekdanielak:
Closing this one. More discussion and debugging will happen in bpmn-js: https://github.com/bpmn-io/bpmn-js/pull/2275

PR #963, opened 2025-03-25
Adjust result sorter to work across keys

Buckwich:
### Checklist

* [ ] **Brief textual description** of the changes present


lib/features/popup-menu/PopupMenuComponent.js
```diff
@@ -88,8 +88,8 @@ export default function PopupMenuComponent(props) {
     return searchFn(searchableEntries, value, {
       keys: [
         'label',
-        'description',
-        'search'
+        'search',
```
nikku, line 91: :star:

nikku:
As per https://github.com/camunda/camunda-modeler/issues/4856#issuecomment-2754045222, let's ensure we can test this isolated in diagram-js, in a big enough, close to real work example.

nikku (changes requested):
Todo to get this merged:

* Test to assert priority (higher up matches beat lower ones)
* Clean up test fixtures to only contain relevant fields (https://github.com/bpmn-io/diagram-js/pull/963/files#diff-1f6fe72e724cb08e8d1db708b6b332f7b5c6ce9cb40686e09023fa4ea74682f0)
* Consider if the real world test should be in search items, or `PopupMenuSpec`. I'd rather have it in popup menu spec to verify the behavior E2E. 
* Name the `searchitems.json` file to be grouped with the spec (cf.  [existing pattern](https://github.com/bpmn-io/bpmn-js/tree/develop/test/spec)). When part of `PopupMenu.spec`, call it `PopupMenu.complex-items.json` or the like.

Buckwich:
I've squashed commits you previously reviewed and dropped the real world example from this branch. I'll rework that properly with your suggestions in #966. also adjusted the test to be simpler and concrete

test/spec/features/search/searchSpec.js
```diff
@@ -144,13 +144,72 @@ describe('features/search', function() {
...
+        search: 'something else A',
+      },
+      {
+        label: 'something else B',
+        search: 'term',
+      },
+    ];
+    const searchItems = (items, term) => search(items, term, {
+      keys: [
+        'search',
```
nikku, line 198: What is confusing here is that we switch keys compared to previous setups. What if we (1) keep keys and items or (2) swap items (input) instead?

test/spec/features/search/searchSpec.js
```diff
@@ -144,13 +144,72 @@ describe('features/search', function() {
...
-    expect(results[1].item).to.eql(items[5]);
-    expect(results[2].item).to.eql(items[0]);
+    expect(results[1].item).to.eql(items[0]);
+    expect(results[2].item).to.eql(items[5]);
     expect(results[3].item).to.eql(items[4]);
     expect(results[4].item).to.eql(items[3]);
   }));
 
 
+  it('should prioritize start of search', inject(function(search) {
```
nikku, line 154:
```suggestion
  it('should prioritize secondary key start of word', inject(function(search) {
```

test/spec/features/search/searchSpec.js
```diff
@@ -144,13 +144,72 @@ describe('features/search', function() {
...
+    // given
+    const items = [
+      {
+        label: 'Invoice Form',
+        search: 'resource form',
+      },
+      {
+        label: 'User task',
+        search: 'form human',
+      } ];
```
nikku, line 165:
```suggestion
      }
    ];
```

nikku: (sorry for that one :see_no_evil:)

test/spec/features/search/searchSpec.js
Buckwich: original idea was to make it clear that i want to test the different order, i was concerned otherwise it could seem that it just takes the original order. i hope i made i more clear now, I think I applied you suggestion

nikku (approved):
Looks good to me, see [this test](https://github.com/bpmn-io/diagram-js/pull/963/files#diff-7dc68f128643d145cf2a06ab4ce54a7cc10388034805b1f6cbe730303853583dR1498) verifying that popup menu works according to your changes.

Updated your changes to be a `feat`. I think it is not a bug fix, but an improvement of the existing search.

Buckwich:
> Updated your changes to be a feat. I think it is not a bug fix, but an improvement of the existing search.

Thanks. I originally choose fix as this was in response to a bug report

PR #966, opened 2025-03-31
Support search keys of string arrays

Buckwich:
Related to https://github.com/camunda/camunda-modeler/issues/4853

Based on https://github.com/bpmn-io/diagram-js/pull/963

### Proposed Changes

This PR changes the search result sort algorithm too allow entering arrays as part of the search items. Main use case is to support the search via keywords. while preparing the entries keywords are added to the search array and can be found. to make the search feel better some adjustments were made to the scoring, eg start of word has now a higher priority.

 
### Visual of search with keywords enabled

https://github.com/user-attachments/assets/4f916175-9413-48dc-8f7e-1280d31cfea5

### Try out

Make sure to close all instances of camunda-modeler

`npx @bpmn-io/sr "camunda/camunda-modeler" -l "bpmn-io/diagram-js#4853-support-keywords-search2"  -c 'npm run dev'`

The new search mostly differs only in Camunda 8 so make sure you are in that version. I think in 7 we don't have the `search` field yet

> At the moment the templates don't have keywords so you won't be able to search for keywords

To test with keywords you need to run [tests](https://github.com/bpmn-io/diagram-js/pull/966/files#diff-7dc68f128643d145cf2a06ab4ce54a7cc10388034805b1f6cbe730303853583dR1612) with added example keywords. Once run you can debug them and play around with the search witht keywords

### Checklist


lib/features/popup-menu/PopupMenuComponent.js
```diff
@@ -88,8 +88,9 @@ export default function PopupMenuComponent(props) {
     return searchFn(searchableEntries, value, {
       keys: [
         'label',
-        'description',
-        'search'
+        'search',
+        'keywords',
```
nikku, line 92: If we start so source `keywords`, then we need test cases for it. However I don't see why we'd need this (and instead rely on `search=string[]` instead.

https://github.com/bpmn-io/bpmn-js-create-append-anything/pull/50#pullrequestreview-2729084390

nikku: What we want to do is to extend the popup menu entry types, so they include `search=string|string[]`.

test/spec/features/search/searchSpec.js
```diff
@@ -144,13 +144,52 @@ describe('features/search', function() {
...
     expect(results[4].item).to.eql(items[3]);
   }));
 
 
+  it('should sort with all keys', inject(function(search) {
+
+    const items = [
+      {
+
+        id: 'resources-replace-form-0',
```
nikku, line 159: I propose we simplify this example to make it easier to understand:

* You test that `f` and `form` matches items in `search`, we don't give up after comparing the label.
* Also needs testing:
  * An entry can be an array
  * Matches are prioritized top -> bottom (an earlier "start of word" match beats a later one)

lib/features/search/search.js
```diff
@@ -22,8 +24,8 @@
  *
  * @template T
  *
- * @param {T[]} items
- * @param {string} pattern
+ * @param {T[]} items Array of objects to be searched
```
nikku, line 27:
```suggestion
 * @param {T[]} items to be searched
```

lib/features/search/search.js
```diff
@@ -22,8 +24,8 @@
...
+ * @param {string} pattern Search pattern, for example, entered by the user
```
nikku, line 28:
```suggestion
 * @param {string} pattern to search for
```

lib/features/search/search.js
```diff
@@ -73,12 +83,32 @@ function matchItem(item, words, keys) {
     matchedWords,
     tokens
   } = keys.reduce((result, key) => {
-    const string = item[ key ];
+    const itemContent = item[ key ]; // could be string|string[]
```
nikku, line 86: Proper typing will catch this, do we have it so this comment is unnecessary?

lib/features/search/search.js
```diff
@@ -73,12 +83,32 @@ function matchItem(item, words, keys) {
...
+
+    if (isArray(itemContent)) {
```
nikku, line 88: Simplification:

```
if (!isArray(itemContent)) {
  itemContent = [ itemContent ];
}
```

Then go ahead and have one flow.

lib/features/search/search.js
Buckwich: i already changed that. (PR was not ready to review yet 😄 )

lib/features/search/search.js
```diff
@@ -20,10 +22,10 @@
 /**
  * Search items by query.
  *
- * @template T
+ * @template [T=Record<string, string | string[]>]
```
Buckwich, line 25: I was looking for something like in TS: `T extends Record...`). Does anyone know how to express that in jsdoc?

lib/features/search/search.js
```diff
@@ -55,15 +57,22 @@ export default function search(items, pattern, options) {
       item,
       tokens
     };
-  }).sort(createResultSorter(keys));
+  });
+
+  const scoredResult = filteredResult.map((result) => {
+    const score = getMaxScore(result, keys);
+    return { ...result, score };
```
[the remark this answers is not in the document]
Buckwich: almost a good suggestion

Buckwich:
some of the existing tests fail now. I'll go through them and check if they are really broken or if I'll just adjust the expected behavior

test/spec/features/popup-menu/PopupMenuSpec.js
```diff
@@ -1586,6 +1587,38 @@ describe('features/popup-menu', function() {
...
+      await expectEntries([ 'GitHub Outbound Connector', 'GitLab Outbound Connector' ]);
+    }));
+
+
+    /**
+     * This test case is only here temporarily to allow experimentation with the new search
+     *
+     * To make it easier to test with sr I added only in this debug commit (will be dropped before merge)
+     */
+    it.only('should allow manual testing', inject(async function(popupMenu) {
```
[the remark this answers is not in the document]
Buckwich: you should read comments

lib/features/search/search.js
nikku: You find your answer [here](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#template).

nikku:
@Buckwich Can we squash the commits on this branch, so they semantically capture what you're doing? Cf. [`DEFINITION_OF_DONE`](https://github.com/bpmn-io/internal-docs/blob/main/development/DEFINITION_OF_DONE.md):

> * available as feature branch on GitHub
>   * contains the cleaned up commit history
>   * commit messages satisfy our [commit message guidelines](https://www.conventionalcommits.org)
>   * one commit on the branch closes the issue via a `Closes #issuenr` statement in the body

test/spec/features/popup-menu/PopupMenu.example-entries.json
```diff
@@ -0,0 +1,1487 @@
...
+    }
+  },
+  {
+    "id": "create-user-task",
+    "label": "User task",
+    "group": {
+      "id": "tasks",
+      "name": "Tasks"
+    },
+    "search": "form human"
```
nikku, line 60: Shouldn't it be this, going into the future? 

```suggestion
    "search": [ "form", "human" ]
```

test/spec/features/popup-menu/PopupMenu.example-entries.json
```diff
@@ -0,0 +1,1487 @@
...
+    "rank": -1
+  },
+  {
+    "id": "create-rule-task",
+    "label": "Business rule task",
+    "group": {
+      "id": "tasks",
+      "name": "Tasks"
+    },
+    "search": "dmn decision"
```
nikku, line 107: Shouldn't it be this going into the future?

```suggestion
    "search": [ "dmn", "decision" ]
```

test/spec/features/popup-menu/PopupMenu.example-entries.json
```diff
@@ -0,0 +1,1487 @@
...
+    "search": "subprocess"
+  },
+  {
+    "id": "create-collapsed-ad-hoc-subprocess",
+    "label": "Ad-hoc sub-process (collapsed)",
+    "group": {
+      "id": "subprocess",
+      "name": "Sub-processes"
+    },
+    "search": "adhoc subprocess"
```
nikku, line 169: Where is this coming from? I don't see it [injected anywhere](https://github.com/camunda/camunda-bpmn-js/blob/main/lib/camunda-cloud/features/popup-menu/CamundaDetailsPopupMenuProvider.js).

test/spec/features/search/searchSpec.js
```diff
@@ -94,10 +94,10 @@ describe('features/search', function() {
       });
 
       // then
-      expect(result[0].tokens).to.have.keys([ 'title', 'description' ]);
-      expect(result[1].tokens).to.have.keys([ 'title', 'description' ]);
+      expect(result[0].tokens).to.have.keys([ 'title0', 'description0' ]);
```
nikku, line 97: What is the contract here? Why are tokens no longer grouped by keys? This is a major breaking change.

test/spec/features/search/searchSpec.js
```diff
@@ -261,10 +261,7 @@ describe('features/search', function() {
 
     // when
     const results = search(items, 'foo', {
-      keys: [
-        'title',
-        'description'
-      ]
+      keys: [ 'title' ]
```
nikku, line 264: How is this the expected behavior? How does `foo` not match `foobar` first (start of word)?

test/spec/features/search/searchSpec.js
```diff
@@ -282,15 +279,15 @@ describe('features/search', function() {
...
-        title: 'foo bar baz'
-      },
       {
         title: 'baz and very long foo bar bar bar\nalso foo bar'
       },
       {
         title: 'foo baz and very long additional text\nalso foo bar'
       },
+      {
+        title: 'foo bar baz'
```
nikku, line 289: Why does `foo bar baz` no longer appear on top, it is the "closest match", also start of word.

test/spec/features/search/searchSpec.js
```diff
@@ -369,40 +366,6 @@ describe('features/search', function() {
   }));
 
 
-  it('should sort alphabetically', inject(function(search) {
```
nikku, line 372: We do no longer search alphabetically? How do we search instead? Let's capture the new behavior in a test case.

test/spec/features/search/searchSpec.js
Buckwich: because I prioritize whole words a bit higher now

test/spec/features/popup-menu/PopupMenu.example-entries.json
Buckwich: we could adjust that in the source, but its not coming from element templates

test/spec/features/search/searchSpec.js
Buckwich: what would you expect the tokens to look like?

lets use the examtle:
```
{
  "title": "Hello"
  "search": [ "An Action", "Something Else"]
} 
```

currently tokens would look like this:
```
{
  title0: tokenized("Hello"),
  search0: tokenized("An Action"),
  search1: tokenized("Something Else"),
}
```

alternative could be 
```
{
  title: [tokenized("Hello")],
  search: [tokenized("An Action"), tokenized("Something Else")],
}
```

or (this doesnot work well with scoring):
```
{
  title: tokenized("Hello"),
  search: tokenized("An Action Something Else"),
}
```
--- 

that `title` becomes an array is due to the simplification, i could adjust that part to only index actual arrays.

We could think about just injecting this custom search in bpmn to not disrupt diagram js users

lib/features/search/search.js
```diff
@@ -73,23 +82,29 @@ function matchItem(item, words, keys) {
...
+    return itemValue.reduce((result, itemString, i) => {
+      const {
+        tokens,
+        matchedWords
+      } = matchString(itemString, words);
+
+      return {
+        tokens: {
+          ...result.tokens,
+          [ key + i]: (result.tokens[ key + i] || []).concat(tokens),
```
nikku, line 100: I wonder if there is a way to "fix this", so that existing tokens are returned, i.e. we could concatenate the tokens (`tokens[key]` is an array when `item[ key ]` is an array).

Background: These tokens are, already today, consumed via the [search pad](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/search-pad/SearchPad.js#L507), and potentially other places, and a these places can use the tokens to render matches, rely on the format. Your proposal breaks the search, I think.

test/spec/features/search/searchSpec.js
Buckwich: I mostly focused on users of camunda-modeler, I thought it wouldn't make sense to sort the results by the alphabetical order of their keywords if the labels match.

test/spec/features/search/searchSpec.js
Buckwich: the reason is that the other entries had 2 occurences of a word match. One could argue that that shouldn't result in a higher score

nikku: > One could argue that that shouldn't result in a higher score

Yes. Otherwise "long content" entries win over these that match exactly; this is what the test originally asserted. :)

test/spec/features/search/searchSpec.js
nikku: Hmm. Alphabetical order gives users something stable, where "entries are equal", I do think that users will benefit from it.

test/spec/features/search/searchSpec.js
nikku: I'd expect: If a search key is an array (new feature), then the tokens become an array (new feature).

You can then decide what to do with these tokens:

* As a `search` consumer, you know what parts are actual arrays.
* If you allow values of a searchable entry to be an array, then you have to handle the results (an array of arrays of tokens) appropriately during rendering

Buckwich:
closing due to too many fixup commits. I can't get them in a sensible order

test/spec/features/popup-menu/PopupMenu.example-entries.json
Buckwich: i think its injected here: https://github.com/bpmn-io/bpmn-js-create-append-anything/blob/main/lib/util/CreateOptionsUtil.js

test/spec/features/search/searchSpec.js
Buckwich: added back in alphabetical sort. but unlikely to occur in desktop modeler due to grouping

test/spec/features/search/searchSpec.js
Buckwich: i'm unable to fix this without breaking the search for ai. just decide which one you prefer. cf. https://github.com/bpmn-io/diagram-js/pull/970/commits/0941843141eb6352f7be6a483595434feeea7255

test/spec/features/search/searchSpec.js
Buckwich: reverted that effect

test/spec/features/search/searchSpec.js
Buckwich: done

test/spec/features/popup-menu/PopupMenu.example-entries.json
```diff
@@ -0,0 +1,1487 @@
+[
```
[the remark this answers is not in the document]
Buckwich: done

PR #967, opened 2025-04-02
Always select first search entry

Buckwich:
Related to https://github.com/camunda/camunda-modeler/issues/4948

### Proposed Changes

always select the first found element

 

https://github.com/user-attachments/assets/65b0f862-ef51-4b9b-9df9-a2471d8d28e3

Steps to try out:

```
npx @bpmn-io/sr "bpmn-io/bpmn-js" -l "bpmn-io/diagram-js#4948-focus-highest-rank-during-search"  -c 'npm start'
```

### Checklist

### DIFF
--- a/lib/features/popup-menu/PopupMenuComponent.js
+++ b/lib/features/popup-menu/PopupMenuComponent.js
@@ -71,21 +71,21 @@ export default function PopupMenuComponent(props) {
     return originalEntries.length > 5;
   }, [ search, originalEntries ]);
 
-  const [ value, setValue ] = useState('');
+  const [ searchValue, setSearchValue ] = useState('');
 
-  const filterEntries = useCallback((originalEntries, value) => {
+  const filterEntries = useCallback((originalEntries, searchValue) => {
 
     if (!searchable) {
       return originalEntries;
     }
 
-    if (!value.trim()) {
+    if (!searchValue.trim()) {
       return originalEntries.filter(({ rank = 0 }) => rank >= 0);
     }
 
     const searchableEntries = originalEntries.filter(({ searchable }) => searchable !== false);
 
-    return searchFn(searchableEntries, value, {
+    return searchFn(searchableEntries, searchValue, {
       keys: [
         'label',
         'description',
@@ -94,23 +94,21 @@ export default function PopupMenuComponent(props) {
     }).map(({ item }) => item);
   }, [ searchable ]);
 
-  const [ entries, setEntries ] = useState(filterEntries(originalEntries, value));
+  const [ entries, setEntries ] = useState(filterEntries(originalEntries, searchValue));
   const [ selectedEntry, setSelectedEntry ] = useState(entries[0]);
 
   const updateEntries = useCallback((newEntries) => {
 
-    // select first entry if non is selected
-    if (!selectedEntry || !newEntries.includes(selectedEntry)) {
-      setSelectedEntry(newEntries[0]);
-    }
+    // always select first
+    setSelectedEntry(newEntries[0]);
 
     setEntries(newEntries);
   }, [ selectedEntry, setEntries, setSelectedEntry ]);
 
-  // filter entries on value change
+  // filter entries on search value change
   useEffect(() => {
-    updateEntries(filterEntries(originalEntries, value));
-  }, [ value, originalEntries ]);
+    updateEntries(filterEntries(originalEntries, searchValue));
+  }, [ searchValue, originalEntries ]);
 
   // handle keyboard seleciton
   const keyboardSelect = useCallback(direction => {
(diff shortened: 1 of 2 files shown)


test/spec/features/popup-menu/PopupMenuComponentSpec.js
```diff
@@ -415,6 +416,41 @@ describe('features/popup-menu - <PopupMenu>', function() {
       expect(domQuery('.selected', container).textContent).to.eql('Entry 3');
     });
 
+    it('should always select first', async function() {
```
nikku, line 419:
```suggestion

    it('should always select first', async function() {
```

test/spec/features/popup-menu/PopupMenuComponentSpec.js
```diff
@@ -415,6 +416,41 @@ describe('features/popup-menu - <PopupMenu>', function() {
...
+      this.timeout(5000);
+      const entriess = [
```
nikku, line 421: Typo, should be `entries`.

test/spec/features/popup-menu/PopupMenuComponentSpec.js
```diff
@@ -415,6 +416,41 @@ describe('features/popup-menu - <PopupMenu>', function() {
...
+        { id: '1', label: 'Hugging' },
+        { id: '2', label: 'Human' },
+        { id: '3', label: 'World' },
+        { id: '4', label: 'Entry 4' },
+        { id: '5', label: 'Entry 5' },
+        { id: '6', label: 'Entry 6' }
+      ];
+
+      // given
```
nikku, line 430: Please also add `when` and `then` to the tests, cf. [good unit tests](https://gist.github.com/nikku/1dc2edb553565238c7bf).

nikku (commented):
Good change on `value -> searchValue`.

UX wise I think this is _exactly_ what we need:

capture GiWgJh_optimized

Let's follow-up on the code changes and get this merged.

test/spec/features/popup-menu/PopupMenuComponentSpec.js
Buckwich: not quite sure about the firs expect block, I would consider it part of given, I just use them to assert the start is what i want. Do you have suggestions how to handle this?

test/spec/features/popup-menu/PopupMenuComponentSpec.js
Buckwich: i didn't really like to overrite the other entries. So it was intentional to be different, but yeah not the best name for that. changed to `otherEntries`

test/spec/features/popup-menu/PopupMenuComponentSpec.js
```diff
@@ -416,9 +416,12 @@ describe('features/popup-menu - <PopupMenu>', function() {
       expect(domQuery('.selected', container).textContent).to.eql('Entry 3');
     });
 
+
     it('should always select first', async function() {
-      this.timeout(5000);
-      const entriess = [
+      this.timeout(5000); // required as each trigger takes 500ms
```
nikku, line 421: You could check if [this](https://github.com/bpmn-io/diagram-js/blob/883efc6a8af88501509a1d8e836ca4605eb9dcce/test/spec/features/popup-menu/PopupMenuComponentSpec.js#L849) can be replaced with the `waitFor` helper, to speed up the test.

Cf. https://github.com/bpmn-io/diagram-js/blob/a74d522f01c2eb434e766a251f070757dc345626/test/spec/features/popup-menu/PopupMenuSpec.js#L649

nikku:
Converted to draft @Buckwich, please mark as ready for review + re-request review once done.

test/spec/features/popup-menu/PopupMenuComponentSpec.js
Buckwich: i can't get it to work. I think the ui is too slow to update, I also tried to rely on spying on the onselect method, but that did also not work. As you can see in my recording there seems to be some delay between input and showing the new search, probably something to investigate aswell. 
This test takes slightly more than 2s, would you be fine with keeping it as is?

nikku: I'm having a look. Thanks!

Buckwich: i moved the test to the feature level. there i had more helper functions and it works now

test/spec/features/popup-menu/PopupMenuComponentSpec.js
Buckwich: and back to entries ...

Buckwich:
Sorry for that mess, I've already squashed my commits. Hope this is now ready to review

nikku:
What I think about when I think "good tests" is something like this: https://github.com/bpmn-io/diagram-js/pull/968.

Simple and clear APIs that allow me to build my tests, to check various scenarios. [KISS](https://en.wikipedia.org/wiki/KISS_principle) style.

nikku (approved):
Re-ordered the test (after assertions that validate search behavior). Looks good to me.

PR #979, opened 2025-06-18
test(bendpoints): verifying that side effect do not occur

AdamBanham:
### Proposed Changes

Adds testing for the issue presented in #965.

But I cannot recapture all reported behaviour in #965.
Testing on root svg element (which fails on develop without the fix from #978) works. 
But for element.out on shapes I was unable to trigger the same occurrences, perhaps due to the mock setup.
Even when I copy the exact workings from the  codepen :/

If maintainers could take a look at how I am triggering the interaction events for shapes and ensure that my approach is suitable within the mock setup that would be great.

Related to #965 
Related to #978

### Checklist

* [X] **Brief textual description** of the changes present
* [X] Related issue linked via `Closes {LINK_TO_ISSUE}` or `Related to {LINK_TO_ISSUE}`

### DIFF
--- a/test/spec/features/bendpoints/BendpointsSpec.js
+++ b/test/spec/features/bendpoints/BendpointsSpec.js
@@ -7,6 +7,9 @@ import modelingModule from 'lib/features/modeling';
 import bendpointsModule from 'lib/features/bendpoints';
 import rulesModule from './rules';
 import interactionModule from 'lib/features/interaction-events';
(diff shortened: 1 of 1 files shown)


nikku:
I'm closing this PR in favor of https://github.com/bpmn-io/diagram-js/pull/1089 - I can confirm the original bug reported is fixed, E2E. Thanks @AdamBanham again for your bug report and your contribution :star:.

PR #1021, opened 2026-04-02
feat: show outline around lasso-selected elements

ev-codes:
### Proposed Changes

related to https://github.com/bpmn-io/bpmn-js/issues/173

Add a selection outline to elements enclosed by the lasso.

Try out using:

`npx @bpmn-io/sr bpmn-io/bpmn-js -l bpmn-io/diagram-js#bpmn-js-173-show-select-in-lasso`

Demo in `bpmn-js`:

https://github.com/user-attachments/assets/2f61096c-2826-4bc1-888a-ea7fe600256b

Demo in `dmn-js`:

https://github.com/user-attachments/assets/3c40ec1e-3921-49da-b5ae-14a757af4dbf

### Checklist

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -146,7 +146,8 @@
   display: none;
 }
 
-.djs-multi-select .djs-element.selected .djs-outline {
+.djs-multi-select .djs-element.selected .djs-outline,
+.djs-dragging-active-lasso .djs-element.selected .djs-outline {
   stroke: var(--element-selected-outline-secondary-stroke-color);
   display: block;
 }
--- a/lib/features/lasso-tool/LassoTool.js
+++ b/lib/features/lasso-tool/LassoTool.js
@@ -27,6 +27,15 @@ import {
  *
  * @typedef {import('../../util/Types').Rect} Rect
  * @typedef {import('../../model/Types').Element} Element
+ *
+ * @typedef { {
+ *   add?: boolean,
+ *   bbox?: Rect,
+ *   marked?: Set<Element>,
+ *   lastMarked?: Set<Element>,
+ *   initialMarked?: Set<Element>,
+ *   frame?: SVGElement
+ * } } LassoContext
  */
 
 var LASSO_TOOL_CURSOR = 'crosshair';
@@ -61,6 +70,9 @@ export default function LassoTool(
   */
   var visuals = {
 
+    /**
+     * @param {LassoContext} context
+     */
     create: function(context) {
       var container = canvas.getActiveLayer();
       var frame = context.frame = svgCreate('rect');
@@ -80,6 +92,9 @@ export default function LassoTool(
       this.update(context);
     },
 
+    /**
+     * @param {LassoContext} context
+     */
     update: function(context) {
       var frame = context.frame,
           bbox = context.bbox;
@@ -111,6 +126,9 @@ export default function LassoTool(
       }
     },
 
+    /**
+     * @param {LassoContext} context
+     */
     remove: function(context) {
 
       this.update(context);
@@ -164,6 +182,9 @@ export default function LassoTool(
 
   eventBus.on('lasso.end', 0, function(event) {
 
+    /**
+     * @type {LassoContext}
+     */
     var context = event.context;
 
     self._select(Array.from(context.marked));
@@ -171,6 +192,9 @@ export default function LassoTool(
 
   eventBus.on('lasso.start', function(event) {
 
+    /**
+     * @type {LassoContext}
+     */
     var context = event.context;
 
     context.bbox = toBBox(event);
@@ -190,6 +214,9 @@ export default function LassoTool(
 
   eventBus.on('lasso.move', function(event) {
 
+    /**
+     * @type {LassoContext}
+     */
     var context = event.context;
 
     var bbox = context.bbox = toBBox(event);
@@ -206,6 +233,9 @@ export default function LassoTool(
 
   eventBus.on('lasso.cleanup', function(event) {
 
+    /**
+     * @type {LassoContext}
+     */
     var context = event.context;
 
     context.lastMarked = context.marked;
@@ -327,6 +357,11 @@ LassoTool.prototype.isActive = function() {
 };
 
 
+/**
+ * @param {Event} event
+ *
+ * @return {Rect}
+ */
 function toBBox(event) {
 
   var start = {
--- a/lib/features/lasso-tool/LassoTool.js
+++ b/lib/features/lasso-tool/LassoTool.js
@@ -1,6 +1,8 @@
 import { values } from 'min-dash';
 
-import { getEnclosedElements } from '../../util/Elements';
+import {
+  getEnclosedElements
+} from '../../util/Elements';
 
 import {
   hasSecondaryModifier
@@ -9,6 +11,7 @@ import {
 import {
   append as svgAppend,
   attr as svgAttr,
+  classes as svgClasses,
   create as svgCreate,
   remove as svgRemove
 } from 'tiny-svg';
@@ -23,9 +26,13 @@ import {
  * @typedef {import('../tool-manager/ToolManager').default} ToolManager
  *
  * @typedef {import('../../util/Types').Rect} Rect
+ * @typedef {import('../../model/Types').Element} Element
  */
 
 var LASSO_TOOL_CURSOR = 'crosshair';
+
+var MARKER_SELECTED = 'selected';
+var LASSO_ACTIVE_CLS = 'djs-dragging-active-lasso';
 
 /**
  * @param {EventBus} eventBus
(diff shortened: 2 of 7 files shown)


nikku:
Try out via 

```
npx @bpmn-io/sr bpmn-io/bpmn-js -l bpmn-io/diagram-js#bpmn-js-173-show-select-in-lasso
```

lib/features/lasso-tool/LassoTool.js
```diff
@@ -90,6 +98,70 @@ export default function LassoTool(
...
+    var currentlyMarked = context.marked || [];
+
+    forEach(desiredMarked, function(e) {
+      if (currentlyMarked.indexOf(e) === -1) {
+        canvas.addMarker(e, MARKER_SELECTED);
+      }
+    });
+
+    forEach(currentlyMarked, function(e) {
+      if (desiredMarked.indexOf(e) === -1) {
```
nikku, line 141:
```suggestion
      if (!desiredMarked.includes(e)) {
```

lib/features/lasso-tool/LassoTool.js
```diff
@@ -90,6 +98,70 @@ export default function LassoTool(
...
+        canvas.removeMarker(e, MARKER_SELECTED);
+      }
+    });
+
+    svgClasses(canvas.getContainer())[desiredMarked.length > 0 ? 'add' : 'remove'](LASSO_ACTIVE_CLS);
+
+    context.marked = desiredMarked;
```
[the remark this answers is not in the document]
nikku: Valid point - ID is sufficient for both `Canvas#addMarker` ([ref](https://github.com/bpmn-io/diagram-js/blob/develop/lib/core/Canvas.js#L694)) and caching (marked can be a `Record<string, boolan>` or a `Set` as suggested).

lib/features/lasso-tool/LassoTool.js
```diff
@@ -134,6 +208,10 @@ export default function LassoTool(
     visuals.create(context);
 
     context.selection = selection.get();
+    context.marked = context.selection.slice();
+
+    domEvent.bind(document, 'keydown', onKeyEvent);
```
nikku, line 213: Why do we do this? It does not look right.

assets/diagram-js.css
```diff
@@ -151,6 +151,11 @@
   display: block;
 }
 
+.djs-lasso-active .djs-element.selected .djs-outline {
+  stroke: var(--element-selected-outline-secondary-stroke-color);
+  display: block;
```
nikku, line 156: We seem to simply copy the multi-select styling and should merge both blocks.

nikku (commented):
UX wise this works great - I could not break it.

Code wise, I don't understand why we need keyboard binding / unbinding.

lib/features/lasso-tool/LassoTool.js
```diff
@@ -90,6 +98,70 @@ export default function LassoTool(
...
+      }
+    });
+
+    forEach(currentlyMarked, function(e) {
+      if (desiredMarked.indexOf(e) === -1) {
+        canvas.removeMarker(e, MARKER_SELECTED);
+      }
+    });
+
+    svgClasses(canvas.getContainer())[desiredMarked.length > 0 ? 'add' : 'remove'](LASSO_ACTIVE_CLS);
```
nikku, line 146: This is not "lasso active", but "multi-selection active, so the nested selected element has a different color. 

For the lasso tool, an outer selection rect is _always_ active, hence I don't quite understand why we need to distinguish here - I propose we _always_ add the class. 

Generally speaking, `dragging` already adds such classes, we could add a dedicated `djs-dragging-active-lasso`:

Image

lib/features/lasso-tool/LassoTool.js
```diff
@@ -142,13 +220,19 @@ export default function LassoTool(
 
     context.bbox = toBBox(event);
     visuals.update(context);
```
nikku, line 222: We already update the visuals here, why do we add a secondary update of visuals? Does it make sense to move the update into the visuals?

nikku (commented):
@ev-codes please look into my comments - see what applies. Solution works great from the user point of view:

Image

lib/features/lasso-tool/LassoTool.js
ev-codes: Fair question!

**Summary:** to accurately preview what gets selected when the lasso is released. But I actually think we should first [change how original selections are handled by the lasso](https://github.com/bpmn-io/diagram-js/issues/1028) to avoid these keyup/keydown handlers in this PR.

----

The goal of the lasso is to preview to the user what will actually be selected after the lasso tool is released.

Some elements can already be selected just before the lasso tool is enabled (a.k.a., the original selection).

If the user is holding Shift as the lasso tool is released, then the original selection will remain selected. Otherwise, the original selection will not be selected.

As I suggested in an internal Camunda demo, we could align this behaviour with other modeling/diagramming tools on the market. The library would instead check if the Shift key is pressed when the lasso tool is enabled, instead of at the end.

[Changing this behaviour first](https://github.com/bpmn-io/diagram-js/issues/1028) would make this PR much simpler, because we wouldn't need to keep track of keyup/keydown. And at the demo, no one disagreed with this approach.

ev-codes:
Moving this back to draft until I complete this task first: https://github.com/bpmn-io/diagram-js/issues/1028

[See explanation here](https://github.com/bpmn-io/diagram-js/pull/1021#discussion_r3099412789)

lib/features/lasso-tool/LassoTool.js
ev-codes: Good idea, thanks! Implemented 🙂

lib/features/lasso-tool/LassoTool.js
ev-codes: My idea was to keep the more SVG-focused updates and the "which elements should be previewed" updates in different functions.

But I can also see how keeping all visual updates in one place would make more sense. So I moved the contents of `updatePreviewMarkers` into `visuals.update` 🙂

lib/features/lasso-tool/LassoTool.js
ev-codes: Now that [the issue discussed above](https://github.com/bpmn-io/diagram-js/issues/1028) is completed, I removed the unneeded keyup/keydown handlers 🙂

test/spec/features/lasso-tool/LassoToolPreviewSpec.js
```diff
@@ -0,0 +1,361 @@
...
+
+      // when: drag without shift
+      dragging.move(canvasEvent({ x: 430, y: 200 }));
+
+      // then: child1 marker is hidden (no shift = won't be in final selection)
+      expect(canvas.hasMarker(child1, 'selected')).to.be.false;
+
+      // only newly enclosed elements have preview markers
+      expect(canvas.hasMarker(child2, 'selected')).to.be.true;
+      expect(canvas.hasMarker(child3, 'selected')).to.be.true;
```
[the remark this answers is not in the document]
ev-codes: I covered the end-of-lasso behavior in another test in a later version of the PR 🙂

lib/features/lasso-tool/LassoTool.js
nikku: Via https://github.com/bpmn-io/diagram-js/pull/1030, specifically https://github.com/bpmn-io/diagram-js/pull/1030/commits/1c30849354d3672d21f8addde6fc501abea70327 I propose the clear separation - please have a look to see if this makes the workings more readable.

nikku:
Thanks for the update @ev-codes, goes into a great direction.

Can you have a look at https://github.com/bpmn-io/diagram-js/pull/1030 to see if the changes on that branch make sense to you? I'd love to incorporate them, to keep the code base long term maintainable:

* Clearly separate visuals from logic ([ref](https://github.com/bpmn-io/diagram-js/pull/1021#discussion_r3051449036))
* Minor test cleanups
* Not using old (cruft) API
* As a behavioral change to this PR `ESC` / cancel _fully_ restores the pre-lasso state ([ref](https://github.com/bpmn-io/diagram-js/pull/1030/commits/54c3677a30afe9465e40282d6d4341d012a793bb)) - I think that is the intended behavior

What do you think?

ev-codes:
@nikku Thanks, I left my comments on #1030 directly!

Please re-request my review on #1030 when you're ready 🙂

nikku (approved):
Incorporated https://github.com/bpmn-io/diagram-js/pull/1030 into this branch and cleaned up the commit history (squashed the changes). Good stuff :star:

ev-codes:
@nikku thanks! I'll smoke test and wait for CI to finish.

I plan to merge this today.

nikku:
> I plan to merge this today.

All good - I'll do it now.

nikku:
Works integrated in bpmn-js :star:

nikku:
Released as [v15.13.0](https://github.com/bpmn-io/diagram-js/releases/tag/v15.13.0)

Issue #1023, opened 2026-04-08
Make `Text` utility didi-injectable as `textLayouter`


Skaiir:
### What should we do?

Register the `Text` utility (`lib/util/Text.js`) as a didi-injectable service (`textLayouter`)
in the draw module.

Currently, consumers like bpmn-js and dmn-js instantiate `Text` directly in their
`TextRenderer`:

```js
var textUtil = new TextUtil({ style: defaultStyle });
```

Instead, `textLayouter` should be a shared service provided through the DI container,
configured via `config.textLayouter`.

This requires changes in three locations:

- **diagram-js**: register `textLayouter` as a service in the draw module
- **bpmn-js**: update `TextRenderer` to inject `textLayouter` instead of instantiating `Text` directly
- **dmn-js**: same as bpmn-js

### Why should we do it?

The text layout engine cannot currently be swapped or extended without replacing
`TextRenderer` entirely. Making it an injectable service creates a clean extension point:
a consumer can replace `textLayouter` with an alternative implementation (e.g., one
backed by a different measurement strategy) without touching the domain-specific
wrappers in bpmn-js or dmn-js.

nikku:
Closing in favor of #1026 - we want to ship a simple improvement at the core. `textRenderer` is already pluggable.

PR #1024, opened 2026-04-08
Make text layouter didi-injectable

Skaiir:
Related to https://github.com/bpmn-io/diagram-js/issues/1023

### Proposed Changes

Register the `Text` utility as a didi-injectable `textLayouter` service in the draw module,
allowing consumers to swap the text layout engine without replacing `TextRenderer`.

- Move implementation from `lib/util/Text.js` to `lib/draw/TextLayouter.js`
- Register as `textLayouter` in the draw module, configured via `config.textLayouter`
- `lib/util/Text.js` becomes a deprecated re-export to preserve backwards compatibility
  for consumers that import `diagram-js/lib/util/Text` directly (both default and named exports)
- Move tests to `test/spec/draw/TextLayouterSpec.js`

The `TextLayouter` constructor remains unchanged , `$inject` is inert for direct
instantiation, so `new TextLayouter(config)` still works as before.

Small note, from the bpmn and dmn side we will no longer rely on passing styles to the constructor. This was already kind of an anti-pattern for the utility and has function-level bypasses to work around exactly that problem, but it will require a small adjustment to how we call createText. This does not affect backwards compatibility. 

### Checklist


nikku (commented):
Thanks for the PR - I think it makes sense. Two remarks:

* Keep some tests for `util/Text` - we don't want to break userland with this change
* Our test files are `.js`, not `.ts`, please adjust accordingly

Skaiir:
@nikku regarding the tests, I didn't change anything these are just renamed copies because the code was moved, so I moved them as well to preserve collocation.  The `.ts` was already there, but I'll adjust. Do we really need to keep tests in `util/Text` considering `Text.js` is now just a pure re-export?

Skaiir:
@nikku Actually on second look all of the test suites in djs are `.ts`, maybe better fixing this all in one go.

nikku:
> Do we really need to keep tests in util/Text considering Text.js is now just a pure re-export?

My ask is to keep some tests, to verify `Text.js` can still be imported and used.

> @nikku Actually on second look all of the test suites in djs are .ts, maybe better fixing this all in one go.

Our type tests end with `.ts` - our functional tests end with `.js`. Please ensure that functional tests are plain javascript - and type interface tests are `.ts` files.

Skaiir:
Makes sense, I added a small test file that smoke tests the export and the functions you can access. I think that should be enough to prevent accidentally breaking the backward compatible exports. I believe the tests are using the proper extensions as well.

nikku:
Awesome :star: - I'll have another look later today.

nikku:
@Skaiir So this PR makes the Text util available as a service - but how is it supposed to be used by downstream, i.e. [bpmn-js](https://github.com/bpmn-io/bpmn-js/blob/develop/lib/draw/TextRenderer.js#L49)? Every editor "configures" the text layouter with a default style. And this seems to be no longer possible.

Please ensure that the PR contains tests that cover a close to real-world use - and from what I understand our intend is to move our existing text layouting from a utility to the service.

Already done: I [moved the text layouter to a dedicated module](https://github.com/bpmn-io/diagram-js/pull/1024/commits/255443344b379fb8f21a22814c3a48868c530935) - that can be used.

nikku:
> Small note, from the bpmn and dmn side we will no longer rely on passing styles to the constructor. **This was already kind of an anti-pattern for the utility and has function-level bypasses to work around exactly that problem**

Please elaborate how that was an anti-pattern, and how the change will look like. And as written, ensure the [future use is covered by tests](https://github.com/bpmn-io/diagram-js/pull/1024#issuecomment-4215570754). A draft PR in bpmn-js can help to convey the message ("intended future use").

Skaiir:
@nikku 

> How is it supposed to be used by downstream

So downstream configuration, if used through didi can still be done via `config.textLayouter`. But this would change the config shape or require an adapter inside of bpmn and dmn-js. Alternatively, my preferred way of doing this is by simply passing the styles through the function calls. This was already supported by the utility.

> Please elaborate how that was an anti-pattern

Because allowing styles to be both provided through the constructor and functions muddies the class scope:

IMO we should have either accepted that styles are something to be provided (1) in the constructor, in which case one Text() instance is scoped to one kind of text styling, or (2) in the function calls, in which case one Text() instance can be used for all your layouting needs as a stateless utility.

We have something hybrid. We set a base style in the constructor that we can then override inside of the function calls.  In `bpmn-js`'s `TextRenderer`, we fully override the style in some function calls (`textUtil.getDimensions`), and don't in others (`textUtil.createText`). 

In my opinion, giving the utility a role of state management makes it inflexible and (2) is the better approach. We do however need to keep the hybrid way going for backwards compatibility reasons.

Draft PR of the changes before yours: https://github.com/bpmn-io/bpmn-js/pull/2412/changes/0fedac7c93a2275de985a49d6e81b8d7a27b0774, this should clarify what I mean in the usage and give a sense of how this works integrated. Note the required adustment to `textUtil.createText`.

nikku:
Thanks @Skaiir! My previous change broke this PR - I pushed https://github.com/bpmn-io/diagram-js/pull/1024/commits/5de052dffa64510e5ab5477a22c8d72992e121ac and it should now be green again.

nikku:
> In my opinion, giving the utility a role of state management makes it inflexible and (2) is the better approach. We do however need to keep the hybrid way going for backwards compatibility reasons.

Do one thing, and do it right - I agree that style merging could be handled in one place - and we can decide to _deprecate_ style merging in `TextLayouter` and move that to the `TextRenderer` instead. 

If we go that route, let's ensure this is [clearly indicated as a deprecation warning](https://github.com/search?q=repo%3Abpmn-io%2Fdiagram-js%20deprecated&type=code) (if used) and **not** exposed as a configuration option. This allows us to get rid of it in a future release.

nikku:
We could also use the [`measureText` API](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/measureText) directly - shipping without an extra dependency.

Skaiir:
@nikku I just benchmarked this. This is definitely the approach we should take (and integrate into the base text utility directly). It doesn't seem to scale quite as well into the 10k+ nodes but this is an edge case, for realistic scenarios, on initial render, it actually outperforms pretext.

This makes sense, pretext is: intial "heavy" computation of text fragments, caching, and then computing based on external size. It's optimized for scenarios with a lot of resizing. I believe it does use the canvas API underneath to do the fragment measuring calculation anyways.

We'd be naively measuring every time we recompute the size, but in most cases this is probably fine. That makes the implementation very tiny, and justifiable in the core IMO. Should I provide a PR?

In terms of measuring performance, it behaves just like pretext.

nikku:
With this you refer to https://github.com/bpmn-io/diagram-js/pull/1027 I believe?

Skaiir:
@nikku Well I built my own benchmarker, but yes, I can pick up the branch from copilot and finish it. Or build my own, idk. I generally dislike copilot's github drafts they tend to miss a lot of things, I think my current local demo works better already (I didn't have issues with alignment that you seem to show). I'll provide something later this afternoon. 

FYI for getting quickly built prototype branches I recommend using the Claude code cli locally (now that we have it). It knows how to interface with github via the github cli and you can at least direct it / interrupt it as much as you want compared to the gh copilot integration that (1) is definitely dumber and has less context capacity, and (2) is hard to direct because it runs async in the background.

nikku:
@Skaiir [the robot draft](https://github.com/bpmn-io/diagram-js/pull/1027) only proved it works by building on top of the existing `measureText` APIs. Please proceed in whatever way you find most suitable - the direction is clear: We can build on top of core APIs directly, don't need to rely on a dependency that is convoluted (for our case).

Skaiir:
Let's then simplify the scope, a surgical change here that will make our life better for a long time. I'll put this PR in draft to keep it visible and in active context, and close all the drafts when the `canvas.measureText()` impl is live.

nikku:
> FYI for getting quickly built prototype branches I recommend using the Claude code cli locally (now that we have it). It knows how to interface with github via the github cli and you can at least direct it / interrupt it as much as you want compared to the gh copilot integration that (1) is definitely dumber and has less context capacity, and (2) is hard to direct because it runs async in the background.

I'm happy to treat it like a junior developer on my side. Works great for me.

nikku:
Closing in favor of https://github.com/bpmn-io/diagram-js/pull/1027.

PR #1030, opened 2026-04-21
Lasso tool internal improvements

nikku:
### Proposed Changes

Building on top of https://github.com/bpmn-io/diagram-js/pull/1021, this proposes 

* test cleanups 
* clear separation of visuals and logic
* simplification of the overall feature

Related to https://github.com/bpmn-io/diagram-js/pull/1021#discussion_r3051449036

### Checklist

### DIFF
--- a/lib/features/lasso-tool/LassoTool.js
+++ b/lib/features/lasso-tool/LassoTool.js
@@ -27,6 +27,15 @@ import {
  *
  * @typedef {import('../../util/Types').Rect} Rect
  * @typedef {import('../../model/Types').Element} Element
+ *
+ * @typedef { {
+ *   add?: boolean,
+ *   bbox?: Rect,
+ *   marked?: Set<Element>,
+ *   lastMarked?: Set<Element>,
+ *   initialMarked?: Set<Element>,
+ *   frame?: SVGElement
+ * } } LassoContext
  */
 
 var LASSO_TOOL_CURSOR = 'crosshair';
@@ -61,6 +70,9 @@ export default function LassoTool(
   */
   var visuals = {
 
+    /**
+     * @param {LassoContext} context
+     */
     create: function(context) {
       var container = canvas.getActiveLayer();
       var frame = context.frame = svgCreate('rect');
@@ -80,6 +92,9 @@ export default function LassoTool(
       this.update(context);
     },
 
+    /**
+     * @param {LassoContext} context
+     */
     update: function(context) {
       var frame = context.frame,
           bbox = context.bbox;
@@ -111,6 +126,9 @@ export default function LassoTool(
       }
     },
 
+    /**
+     * @param {LassoContext} context
+     */
     remove: function(context) {
 
       this.update(context);
@@ -164,6 +182,9 @@ export default function LassoTool(
 
   eventBus.on('lasso.end', 0, function(event) {
 
+    /**
+     * @type {LassoContext}
+     */
     var context = event.context;
 
     self._select(Array.from(context.marked));
@@ -171,6 +192,9 @@ export default function LassoTool(
 
   eventBus.on('lasso.start', function(event) {
 
+    /**
+     * @type {LassoContext}
+     */
     var context = event.context;
 
     context.bbox = toBBox(event);
@@ -190,6 +214,9 @@ export default function LassoTool(
 
   eventBus.on('lasso.move', function(event) {
 
+    /**
+     * @type {LassoContext}
+     */
     var context = event.context;
 
     var bbox = context.bbox = toBBox(event);
@@ -206,6 +233,9 @@ export default function LassoTool(
 
   eventBus.on('lasso.cleanup', function(event) {
 
+    /**
+     * @type {LassoContext}
+     */
     var context = event.context;
 
     context.lastMarked = context.marked;
@@ -327,6 +357,11 @@ LassoTool.prototype.isActive = function() {
 };
 
 
+/**
+ * @param {Event} event
+ *
+ * @return {Rect}
+ */
 function toBBox(event) {
 
   var start = {
--- a/lib/features/lasso-tool/LassoTool.js
+++ b/lib/features/lasso-tool/LassoTool.js
@@ -1,7 +1,7 @@
-import { forEach, values } from 'min-dash';
+import { values } from 'min-dash';
 
 import {
-  getEnclosedElements as _getEnclosedElements
+  getEnclosedElements
 } from '../../util/Elements';
 
 import {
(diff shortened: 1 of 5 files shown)


lib/features/lasso-tool/LassoTool.js
```diff
@@ -162,57 +182,66 @@ export default function LassoTool(
...
-    context.bbox = toBBox(event);
+    var bbox = context.bbox = toBBox(event);
+
+    var enclosedElements = getEnclosedElements(bbox);
+
+    var marked = context.initialMarked.union(enclosedElements);
+
+    context.lastMarked = context.marked;
+    context.marked = marked;
+
```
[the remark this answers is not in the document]
nikku: Available as per [2024 baseline](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/union). We're in 2026 now.

lib/features/lasso-tool/LassoTool.js
```diff
@@ -58,11 +70,13 @@ export default function LassoTool(
   */
   var visuals = {
 
+    /**
+     * @param {LassoContext} context
+     */
     create: function(context) {
-      var container = canvas.getActiveLayer(),
-          frame;
+      var container = canvas.getActiveLayer();
```
ev-codes, line 77: suggestion: if we're making kinds of variable assignment edits, I think we should also use the block-scoped `let` instead of `var` to avoid accidental variable overrides in future.

WDYT? 

question: In this repo, I notice both `let` and `var` being used. Do you know why?

lib/features/lasso-tool/LassoTool.js
```diff
@@ -72,71 +86,77 @@ export default function LassoTool(
...
+          x: bbox.x,
+          y: bbox.y,
+          width: bbox.width,
+          height: bbox.height
+        });
+      }
 
-      var desiredMarkedSet = new Set(desiredMarked);
-      var currentlyMarked = context.marked || new Set();
+      var currentlyMarked = context.marked,
```
ev-codes, line 111: question: why not use two separate variable assignments here like this PR does at line 78?

lib/features/lasso-tool/LassoTool.js
```diff
@@ -72,71 +86,77 @@ export default function LassoTool(
...
+  }
 
-      canvas.removeMarker(e, MARKER_SELECTED);
-    });
+  /**
+   * @param {Rect} bbox
+   *
+   * @return {Set<Element>}
+   */
+  function getEnclosedElements(bbox) {
```
ev-codes: suggestion: I think this function should be renamed to `getAllEnclosedElements`.

The semantics are slightly different than `util/Elements.getEnclosedElements` - all elements are retrieved from the `elementRegistry` first.

I notice the aliased `_getEnclosedElements` is used separately directly on line 327 as well, so some parts of `LassoTool.js` still need to use this utility with only a few selected elements.

test/spec/features/lasso-tool/LassoToolPreviewSpec.js
```diff
@@ -268,16 +268,18 @@ describe('features/lasso-tool - preview styling', function() {
...
 
-      // given: child1 is pre-selected
+      // given
       selection.select([ child1 ]);
 
       // start lasso after child1's right edge (x=180) to avoid re-enclosing it
       lassoTool.activateLasso(canvasEvent({ x: 190, y: 90 }));
 
-      // when: drag with shift held
+      // when
```
ev-codes, line 277: suggestion: tbh I find this comment style harder to read and overly verbose

When the comment takes up more space than the code, I find it a little distracting 😅 

Is this some convention in the repo that I'm missing?

ev-codes (commented):
Thanks! 

A lot of useful changes in here, but there's a few I'm not sure about. I left a few comments.

lib/features/lasso-tool/LassoTool.js
nikku: We historically use `var` - we should do a one off rewrite to `let` / `const` some day in the future. When `let` creeps in, today, then it is not intentional.

lib/features/lasso-tool/LassoTool.js
nikku: Good point, will also allow us to undo the `_getEnclosedElements` rename.

test/spec/features/lasso-tool/LassoToolPreviewSpec.js
nikku: > suggestion: tbh I find this comment style harder to read and overly verbose

It is important for us to [clearly tell apart given, when, then](https://gist.github.com/nikku/1dc2edb553565238c7bf). This is part of our practice and consistently applied across our vast code base.

Doing stuff like

```
// when - I do this

// when: I do something else
```

Sounds nice at first glance but introduces unnecessary burden on this reading the tests.

nikku: To summarize: These are captions within a test, and not meant to be fiddled with. If your code (below) needs a human readable description, then add a comment. :)

lib/features/lasso-tool/LassoTool.js
nikku: Our goal is to have the code readable - historically we use this block.

However, it becomes much more unreadable when we do nested assignments such as

```
var abc = this_expression,
    def = someThingElse ? otherThing : foo || bar;
```

In the later case it helps (me) to separate the two blocks. We don't enforce a style, but aim to produce compact enough, readable code.

We also don't do prettier - and that is OK. Goal is that humans can read what is written, and that what is written makes sense.

lib/features/lasso-tool/LassoTool.js
nikku: Cf. https://github.com/bpmn-io/diagram-js/pull/1030/changes/BASE..54c3677a30afe9465e40282d6d4341d012a793bb#r3122832549 on separating the assignments.

lib/features/lasso-tool/LassoTool.js
```diff
@@ -162,57 +182,66 @@ export default function LassoTool(
...
 
+    /**
+     * @type {LassoContext}
+     */
     var context = event.context;
 
-    visuals.remove(context);
-    clearPreviewed(context);
+    context.lastMarked = context.marked;
+    context.marked = new Set(selection.get());
```
ev-codes, line 242: note: this diff aims to "ESC / cancel fully restores the pre-lasso state"

After a quick survey of other tooling (Miro, n8n, Excalidraw), none of them appear to offer this behavior. And it wasn't my intention in the original PR.

But I tried it out, and I think it makes sense to add!

question: why did you decide to add this? Did some feedback/observation prompt it, or did it just "feel right" to you?

nikku: > question: why did you decide to add this? Did some feedback/observation prompt it, or did it just "feel right" to you?

You mean why did I decide to restore on `ESC`? Because that is what ESC should do - get me out of the current place, back to the previous place. Works consistently in our tools, independent from how other tools do it, example is our space tool:

capture 1mbQRL_optimized

Backed by our [design principles](https://github.com/bpmn-io/design-principles), specifically [do the hard work to make it simple](https://github.com/bpmn-io/design-principles#do-the-hard-work-to-make-it-simple) - allow users to work where they left off before the operation they canceled with `ESC`.

On top, if you choose the simple implementation that this PR contributes then you'll automatically end up in that UX.

philippfromme: Restoring the previous selection on <kbd>ESC</kbd> would be my expectation, too.

philippfromme: I'm testing the <kbd>ESC</kbd> behavior and for me it doesn't work. When I press <kbd>ESC</kbd> the previous selection is restored but as soon as I let go of the mouse the selection is lost.

chrome_lge19QjXRy

lib/features/lasso-tool/LassoTool.js
nikku: Fixed via https://github.com/bpmn-io/diagram-js/pull/1030/commits/c5986dbe6b0e06d1bc4ba97e02912e0c2ddecc99. Local utility is now called `getAllEnclosedElements` as you suggested.

We could get rid of `LassoTool#select`, if it was not part of the components public API. I decided to not touch it for the moment.

lib/features/lasso-tool/LassoTool.js
nikku: @philippfromme what you refer to may be a side-effect of a dragging bug - mouse up is recorded as a click. To be investigated separately.

PR #1037, opened 2026-05-19
feat: add a placeholder in popup menu search

barmac:
### Proposed Changes

A small discoverability improvement in case the user misses the magnifying glass and a blinking cursor.

image

### Checklist

### DIFF
--- a/lib/features/popup-menu/PopupMenu.js
+++ b/lib/features/popup-menu/PopupMenu.js
@@ -23,6 +23,7 @@ import PopupMenuComponent from './PopupMenuComponent';
  * @typedef {import('../../core/Canvas').default} Canvas
  * @typedef {import('../../core/EventBus').default} EventBus
  * @typedef {import('../search/search').default} search
+ * @typedef {import('../../i18n/translate/translate').default} translate
  *
  * @typedef {import('../../util/Types').Point} Point
  *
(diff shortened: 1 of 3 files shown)


lib/features/popup-menu/PopupMenu.js
```diff
@@ -112,6 +116,8 @@ PopupMenu.prototype._render = function() {
     options
   } = this._current;
 
+  const searchPlaceholder = this._translate('Search');
```
nikku, line 119: Consider: Should this be static text or determined by context? 

```
Search
Filter elements...
Type "Gateway"...
```

Are options that could be provided user-land.

Also consider: Is this search or filtering?

jarekdanielak:
I agree with Nico that this should be configurable.

lib/features/popup-menu/PopupMenu.js
barmac: Yes, we could extend the API to make the placeholder configurable.

On search vs. filter: To my understanding, I use a filter to limit the options, but not necessarily select one. In the popup menu, I search a concrete item that I am interested in selecting. So I'd rather call it search.

barmac:
Moved to draft for further rework

barmac:
It seems that we may not need that.

Issue #1056, opened 2026-07-01
Popup menu entries lose their `title` attribute (and accessible name) since 15.15.0


nithinssabu:
## Description
Since diagram-js 15.15.0, popup menu entries (`<li>` items rendered by `PopupMenuItem.js`) that only specify `entry.label` (not `entry.title`) no longer get a `title` attribute in the rendered DOM. This is a regression from 15.14.0, which had a fallback to `entry.label`.

## Root cause
In `lib/features/popup-menu/PopupMenuItem.js`:
- **15.14.0**: `title=${ entry.title || entry.label }`
- **15.15.0+**: `title=${ entry.title }` (fallback dropped)

For comparison, the sibling component `PopupMenuHeader.js` (added in the same 15.15.0 multi-step-navigation release) correctly does `title=${ entry.title || entry.label }` — suggesting the fallback was simply missed on `PopupMenuItem.js` during that refactor.

## Impact
Most popup-menu entry providers (e.g. bpmn-js's append/create/replace menus) only set `.label`, not `.title`, on their entries. As a result, since 15.15.0, these entries render with no `title` attribute at all.

This is a real accessibility regression, not just cosmetic: per the WAI-ARIA spec, the `listitem` role only supports "Name from: author" (`title`/`aria-label`/`aria-labelledby`) — it does *not* support "name from content". Without `title`, these `<li>` elements have no accessible name in the accessibility tree, even though they visually display text.

This also breaks accessibility-first testing patterns, e.g. Playwright's `getByRole('listitem', { name: 'User task' })` no longer matches these entries.

## Reproduction
1. Create a bpmn-js modeler with a `bpmn:StartEvent`.
2. Select the start event and trigger the append popup menu (e.g. via `bpmn-js-create-append-anything`).
3. Inspect the rendered `<li data-id="append-user-task">` — in diagram-js <= 15.14.0 it has `title="User task"`; in >= 15.15.0 it has no `title` attribute.

Verified with a clean A/B comparison: identical popup-menu code, only diagram-js/bpmn-js versions differed.

## Suggested fix
Restore the fallback in `PopupMenuItem.js`:
```diff
- title=${ entry.title }
+ title=${ entry.title || entry.label }
```

## Versions
- Regression introduced: diagram-js 15.15.0
- Still present: diagram-js 15.18.0 (latest checked)
- bpmn-js versions affected: 18.17.0+ (any version bundling diagram-js >= 15.15.0)

[...]

PR #1059, opened 2026-07-03
Improve popup accessibility

### LINKED ISSUE #735: Improve the aural interface of the element template chooser

nikku:
### Proposed Changes

This PR explores improving the accessibility of our popup menu:

* Turns it into a real combobox (with search) / listbox (no search)
* Removes past focus hack
* Works in two dedicated modes (still)
  * Keyboard navigation - shows documentation link not inline, but on the bottom of the box (we'll otherwise not comply with `aria` specification - list items cannot contain focusable elements)
  * Mouse navigation - documentation link is "in focus", on the individual item
* Keyboard navigation is _fixed_. No longer endless tabbing through the elements - a combo-box has _one_ keyboard focus, and I can tab out of it easily

capture U59Jjd_optimized

Relates to https://github.com/bpmn-io/diagram-js/issues/1056 - just a proper fix that _truely_ improves the components accessibility.

Closes https://github.com/bpmn-io/diagram-js/issues/735

### Try out via

```
npx @bpmn-io/sr bpmn-io/bpmn-js-create-append-anything -l bpmn-io/diagram-js#improve-popup-accessibility
```

### Checklist

### DIFF
--- a/test/spec/features/popup-menu/PopupMenuSpec.js
+++ b/test/spec/features/popup-menu/PopupMenuSpec.js
@@ -3189,6 +3189,58 @@ describe('features/popup-menu - integration', function() {
 
   }));
 
+
+  // accessibility playground
+  //
+  // add `.only` to this test and run `npm run dev` to open a rich popup menu
+  // for manual inspection (DevTools > Accessibility pane and/or a screen
+  // reader). Covers the #1056 rework:
+  //
+  //   - entries are role="option", named from their visible label
+  //   - hovering an entry reveals the mouse-only (aria-hidden) docs icon
+  //   - the actionable "Open entry documentation" link sits in the footer and
+  //     reflects the active entry (keyboard/AT reachable via Tab)
+  //   - > 5 entries render a search field (role="combobox") driving the list
+  //     (role="listbox") via aria-activedescendant
+  it('should render accessibility playground', inject(function(popupMenu) {
+
+    // given
+    const tasksGroup = { id: 'tasks', name: 'Tasks' };
+    const gatewaysGroup = { id: 'gateways', name: 'Gateways' };
+
+    const entries = entrySet([
+      { id: 'task', label: 'Task', group: tasksGroup, className: 'bpmn-icon-task',
+        documentationRef: 'https://docs.camunda.io/docs/components/modeler/bpmn/tasks/' },
+      { id: 'service-task', label: 'Service Task', group: tasksGroup, className: 'bpmn-icon-service',
+        description: 'Invokes a service, e.g. a REST endpoint',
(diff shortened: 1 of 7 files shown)


test/spec/features/popup-menu/PopupMenuSpec.js
```diff
@@ -2959,6 +2959,59 @@ describe('features/popup-menu - integration', function() {
...
 
+
+  // accessibility playground
+  //
+  // add `.only` to this test and run `npm run dev` to open a rich popup menu
+  // for manual inspection (DevTools > Accessibility pane and/or a screen
+  // reader). Covers the #1056 rework:
+  //
+  //   - entries are role="option", named from their visible label
+  //   - entries with docs expose aria-description="Further help available"
```
jarekdanielak:
```suggestion
```

jarekdanielak:
I noticed this PR sneaks quite a few excessive comments in CSS and tests. @nikku are you aware and does it work for you? For testing, it's actually against our rules:

> Tests should be self-explanatory. Do not add explanatory comments beyond given/when/then structure. If you need to add comments, consider refactoring the test to be more clear.

cf. https://github.com/bpmn-io/internal-docs/blob/main/development/ai-prompts/claude/rules/testing.md

nikku:
Adjusted some names for clarity, reduced excessive comments. Agreed the later were not necessary.

nikku:
Consideration for the ID generation: instead of the module-level `new Ids('djs-popup')` singleton + `useMemo(() => ids.next(), [])`, we could use Preact's `useId()` (already re-exported from `@bpmn-io/diagram-js-ui`):

```js
import { useId } from '../../ui/index.js';

const idPrefix = useId();
```

This drops the `IdGenerator` import, the shared singleton, and the `useMemo`. It's also more correct:

- `useMemo` gives no stability guarantee (it's a caching hint, not identity) — `useId` is purpose-built for stable per-instance ids.
- `ids.next()` mutates shared module state during render, which isn't render-/StrictMode-safe; `useId` is.
- Output (`P0-0`) is selector-safe, and the existing tests stay green.

Only trade-off is losing the readable `djs-popup` prefix in the DOM — keepable via `` `djs-popup-${useId()}` `` if desired. Non-blocking, just for consideration.

jarekdanielak:
Sometimes the documentation icon (in the footer) is not shown when navigating with keyboard. Cannot reliably reproduce, tho.

image

jarekdanielak:
Works well aside from the above comment. The "tab" behavior is the biggest improvement IMO, worth the hustle.

nikku:
Thanks for the review :star: - let's merge then, and iterate from there.

PR #1060, opened 2026-07-03
Popup menu bug fixes

jarekdanielak:
Related to https://github.com/camunda/camunda-modeler/issues/6027

### Ignore mouse when navigating with keyboard

Prevents a bug happening when you navigate down on the last entry, and you land on the entry which is hovered by mouse, instead of the first one in the list.

Before

djs-popup-to-first-bug

After

djs-popup-to-first-fix

### Restore focus after drilling down into an entry

Allows for keyboard navigation after clicking into an entry with steps. Was not possible before.

### Checklist

### DIFF
--- a/lib/features/popup-menu/PopupMenuComponent.js
+++ b/lib/features/popup-menu/PopupMenuComponent.js
@@ -268,6 +268,7 @@ export default function PopupMenuComponent(props) {
       position=${ position }
       width=${ width }
       scale=${ scale }
+      navigationStack=${ navigationStack }
     >
       ${ displayHeader && html`
         <${PopupMenuHeader}
@@ -330,6 +331,7 @@ function PopupMenuWrapper(props) {
     onKeyup,
     className,
     children,
+    navigationStack,
     position: positionGetter
   } = props;
 
@@ -348,7 +350,6 @@ function PopupMenuWrapper(props) {
     popupEl.style.top = `${position.y}px`;
   }, [ popupRef.current, positionGetter ]);
 
-  // initial focus
   useLayoutEffect(() => {
     const popupEl = popupRef.current;
 
@@ -359,7 +360,7 @@ function PopupMenuWrapper(props) {
     const inputEl = popupEl.querySelector('input');
 
     (inputEl || popupEl).focus();
-  }, []);
+  }, [ navigationStack ]);
 
   // global <Escape> / blur handlers
   useEffect(() => {
--- a/lib/features/popup-menu/PopupMenuList.js
+++ b/lib/features/popup-menu/PopupMenuList.js
@@ -37,8 +37,6 @@ export default function PopupMenuList(props) {
       return;
     }
 
-    keyboardSelection.current = false;
-
     const containerEl = resultsRef.current;
 
     if (!containerEl)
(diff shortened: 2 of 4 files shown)


nikku:
This seems to do many things at once - some we're not even sure if we want them. Can we have separate PRs for the fixes? :pray:

nikku:
> Ignore mouse when navigating with keyboard

Are we fixing an edge case here? I'd assume that the moment mouse hovers it "takes control" - you either work with the keyboard or take over control with the mouse.

jarekdanielak:
I took https://github.com/bpmn-io/diagram-js/pull/1072 out as the "fix" contains an opinionated UI change. The rest of the commits in this PR seem to be quite straight-forward.

jarekdanielak:
> > Ignore mouse when navigating with keyboard
> 
> Are we fixing an edge case here? I'd assume that the moment mouse hovers it "takes control" - you either work with the keyboard or take over control with the mouse.

It's an edge case in some sense - only happens if you have the cursor hovering somewhere over the menu, and you switch to navigating with keyboard.

When navigating with keyboard, I expect the mouse position to always be ignored.

nikku:
> When navigating with keyboard, I expect the mouse position to always be ignored.

So what do we "loose" when we go that route? I'd expect, UX wise:

* Mouse is a navigator - you hover with mouse, you pre-select
* Keyboard accounts for current selection - you arrow up/down again, that pics up from mouse selected element

jarekdanielak:
Both your assumptions hold now.

The bug is: I navigate with keyboard, go "arrow down" from the last item on the list. I expect to land on the first item, but instead I land where my mouse cursor is.

nikku (commented):
I propose we merge one contribution at a time - https://github.com/bpmn-io/diagram-js/pull/1074 flips around "tracks keyboard adjustments" and ends up in a simpler implementation.

It also adds the tests in the right place - PopupMenuSpec - so that we can throw away the internals at some point and the tests still pass. What do you think?

jarekdanielak:
Agreed. I'll rework this on top of the now merged https://github.com/bpmn-io/diagram-js/pull/1074

jarekdanielak:
Rebased on the latest develop. Updated the top comment to reflect the two fixes that are left in this PR.

lib/features/popup-menu/PopupMenuComponent.js
```diff
@@ -359,7 +360,7 @@ function PopupMenuWrapper(props) {
     const inputEl = popupEl.querySelector('input');
 
     (inputEl || popupEl).focus();
-  }, []);
+  }, [ navigationStack ]);
```
nikku, line 363: I understand this fix - I don't understand the other fix. :)

jarekdanielak: Do an experiment:

1. Open popup and position your cursor somewhere in the middle of the list.
2. Navigating with keyboard, go all the way to the end of the list, and one more.
3. Instead of going to the first element, you land on the one that you left your cursor over.

diagram-how-to-bug

jarekdanielak: Now if "the other fix" is applied, you'd land in the first item in the list.

nikku: What does "the other fix" do? Why does it work?

nikku: The first fix we can merge. It is clear what the bug was that we fixed.

jarekdanielak: Root cause: when the list scrolls all the way up on keyboard navigation, it fires `mouseenter` on an entry that ends up under the pointer, stealing the navigation from the keyboard.

jarekdanielak: Added a comment (https://github.com/bpmn-io/diagram-js/pull/1060/commits/56d0edcd9d46016aa4b5ccea06c709662e444b5d) to clarify.

nikku (approved):
Thanks for your clarification in code. I also put that into the commit messages. Verified it works ✔️

PR #1072, opened 2026-07-06
fix(PopupMenu): position doc icon top right

jarekdanielak:
Related to https://github.com/camunda/camunda-modeler/issues/6027

### Proposed Changes

Consistently position the doc icon in the top right corner of an entry, or inline with the chevron if one-line.

image

image

image

### Checklist

* [ ] Pull request __establishes context__

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -681,10 +681,32 @@
   text-decoration: none;
 }
 
-.djs-popup-entry-chevron {
+.djs-popup-entry-actions {
+  container-type: size;
+  position: relative;
+  align-self: stretch;
+  width: 34px;
+  margin-left: 6px;
+}
+
+.djs-popup-entry-actions-inner {
+  position: absolute;
+  inset: 0;
   display: flex;
   align-items: center;
-  margin-left: 6px;
+  justify-content: center;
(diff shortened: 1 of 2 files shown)


nikku:
@jarekdanielak I'm not actually sure this is an improvement. It conflicts with the `>` arrow. I'd not "micro-optimize" here, given that we discuss resizing of the replace menu anyway.

Unrelated: What I suggested earlier was to replace it with a `?` - an icon that (1) fits more into the entry, (2) reflects "help available" better. We can consider that to make the overall UI more readable.

jarekdanielak:
Even if we resize the menu, there is always a case of an entry that's on the edge, so the [related problem](https://github.com/camunda/camunda-modeler/issues/6027) will occur.

My argument against an `?` icon was that the current icon is the best for describing what it does - opens an external source. `?` looks like a tooltip/local help for me.

The improvement that I see here is in consistency - "open docs" icon will always be in the same place for every entry, regardless of how long the title is. WDYT?

jarekdanielak:
Closing as this received mixed feedback, cf. https://camunda.slack.com/archives/C0ADUDNLJJD/p1783521635462279

PR #1076, opened 2026-07-08
Correct popup menu help link not triggering via `ENTER`

nikku:
### Proposed Changes

Previously, we'd still handle ENTER in the popup menu, trigger
the respective (selected) entry, regardless of whether focus was
on the documentation link in the footer.

Related to #1059 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to [diagram-js](https://github.com/bpmn-io/diagram-js) are d
 
 _**Note:** Yet to be released changes appear here._
 
+## 15.20.1
+
+ * `FIX`: allow triggering popup menu documentation link via `ENTER` ([#1076](https://github.com/bpmn-io/diagram-js/pull/1076))
+
 ## 15.20.0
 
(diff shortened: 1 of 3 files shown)


jarekdanielak (changes requested):
I cannot see the icon when navigating via keyboard. Testing on:

```
npx @bpmn-io/sr -l bpmn-io/diagram-js#fix-jelp-link bpmn-io/bpmn-js-create-append-anything
```

Image

nikku:
> I cannot see the icon when navigating via keyboard. Testing on:

You'll not see the footer documentation link when navigating via mouse - you'll see it when the keyboard has selected an entry that has the icon and _NO_ mouse focus exists.

So the bug you're seeing is: You hovered with the mouse - continue to navigate with keyboard - and wonder where the footer link is. Remove the mouse and it is shown (again).

The alternative is that we'd show the entry twice - 

capture p0d2lW_optimized

jarekdanielak:
Got it! It's slightly confusing for me, as a "hybrid" mouse and keyboard user, that we introduced a state where an entry is selected, but no documentation icon is shown, but I guess it makes sense overall.

jarekdanielak (approved):

nikku:
We can collect feedback and follow-up. You provided one data point :slightly_smiling_face:

PR #1082, opened 2026-07-16
feat: add tooltip with title and shortcut on palette entries

AlekseyManetov:
### Proposed Changes

Related to https://github.com/camunda/camunda-modeler/issues/6011

Add a styled hover tooltip to core palette entries showing the entry title and, when set, a keyboard shortcut. Since entry titles are now rendered in tooltip, native html `title` property replaced  with `aria-label` (+ `role="button"`), to avoid additional native browser tooltip.

Follow up `bpmn-js` PR with providing shortcuts to the entries - https://github.com/bpmn-io/bpmn-js/pull/2465

https://github.com/user-attachments/assets/17baf956-6de9-4828-928b-27085f350909

**Try out:**
```
npx @bpmn-io/sr bpmn-io/bpmn-js#add-pallet-tooltips -l bpmn-io/diagram-js#add-palette-tooltips
```

### Checklist

  * [x] __Steps to try out__, i.e. [using the `@bpmn-io/sr` tool](https://github.com/bpmn-io/sr)

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -100,6 +100,13 @@
   --tooltip-error-background-color: var(--color-red-360-100-97);
   --tooltip-error-border-color: var(--color-red-360-100-45);
   --tooltip-error-color: var(--color-red-360-100-45);
+
+  --hover-tooltip-font-family: "IBM Plex Sans", sans-serif;
+  --hover-tooltip-background-color: var(--color-grey-225-10-15);
+  --hover-tooltip-color: var(--color-white);
+  --hover-tooltip-shadow-color: var(--color-black-opacity-10);
+
+  --palette-tooltip-shortcut-background-color: var(--color-grey-225-10-35);
 }
 
 /**
@@ -482,6 +489,49 @@
 
 .djs-palette.open .djs-palette-toggle {
   display: none;
+}
+
+.djs-hover-tooltip {
+  position: fixed;
+  z-index: 1000;
+
+  display: flex;
+  align-items: center;
+  gap: 6px;
+
+  padding: 6px 10px;
+  border-radius: 4px;
+
+  background-color: var(--hover-tooltip-background-color);
+  color: var(--hover-tooltip-color);
+
+  font-family: var(--hover-tooltip-font-family);
+  font-size: 12px;
+  line-height: 1.2;
+  white-space: nowrap;
+
+  pointer-events: none;
+  box-shadow: 0 1px 4px var(--hover-tooltip-shadow-color);
+}
+
+.djs-hover-tooltip--right {
+  transform: translateY(-50%);
+}
+
+.djs-hover-tooltip--top {
+  transform: translate(-50%, -100%);
+}
+
+.djs-palette-tooltip-shortcut {
+  padding: 1px 5px;
+
+  border-radius: 3px;
+
+  background-color: var(--palette-tooltip-shortcut-background-color);
+
+  font-family: monospace;
+  font-size: 11px;
+  line-height: 1.2;
 }
 
 /**
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -109,6 +109,15 @@
   --tooltip-error-background-color: var(--color-red-360-100-97);
   --tooltip-error-border-color: var(--color-red-360-100-45);
   --tooltip-error-color: var(--color-red-360-100-45);
+
+  --hover-tooltip-font-family: "IBM Plex Sans", sans-serif;
+  --hover-tooltip-font-size: 12px;
+  --hover-tooltip-secondary-font-size: 11px;
+  --hover-tooltip-background-color: var(--color-grey-225-10-15);
+  --hover-tooltip-color: var(--color-white);
+  --hover-tooltip-shadow-color: var(--color-black-opacity-10);
+
+  --palette-tooltip-shortcut-background-color: var(--color-grey-225-10-35);
 }
 
 /**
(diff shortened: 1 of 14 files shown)


lib/features/palette/Palette.js
```diff
@@ -262,7 +262,8 @@ Palette.prototype._update = function() {
       domAttr(control, 'data-action', id);
 
       if (entry.title) {
-        domAttr(control, 'title', entry.title);
+        domAttr(control, 'aria-label', entry.title);
+        domAttr(control, 'role', 'button');
       }
```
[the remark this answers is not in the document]
jarekdanielak: @AlekseyManetov Worth verifying

lib/features/palette/PaletteTooltip.js
barmac, line 1: [q] Is a tooltip a special thing for the palette, or rather a general utility? Let's say I were to add tooltips to the context pad items (like we do in improved-canvas). Do you see any common way we could make it work?

AlekseyManetov: Yes, I think we can extract some common parts so we're able to reuse tooltip in other places. I'll handle it.

AlekseyManetov: I've reworked it to be general utility now, not palette-specific. I extracted a reusable `hoverTooltip` that handles hover + delay + positioning + render lifecycle and takes opaque content — consumers just supply what to show. `PaletteTooltip` is a thin wrapper; context pad would add its own thin consumer via `hoverTooltip.add({ container, selector, getContent, getPosition })` with a different placement. One mechanism, thin per-feature consumers.

Naming was the tricky part, because diagram-js already has a `tooltips` component — but it's a genuinely different concept: canvas-coordinate annotations that are pinned to a diagram position and pan/zoom with the canvas, added/removed imperatively (used for ModelingFeedback's rejection messages). What we need here is the opposite — a screen-fixed, hover-triggered hint on UI controls that must not move with the canvas. Different coordinate space, lifecycle, and trigger, so reusing tooltips would mean mixing two models into one service. 

Ideally I'd rename the pre-existing tooltips to something like `canvasHint`, but that's a breaking change. So I went with `hoverTooltip` for now — not fully sold on it(also weighed uiTooltip, controlTooltip).

@barmac thoughts on the approach, and a better name?

lib/features/palette/Palette.js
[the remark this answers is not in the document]
AlekseyManetov: Keyboard navigation of the palette wasn't available before this PR(ref. https://github.com/camunda/camunda-modeler/issues/5687) and isn't changed by it — entries were never in the tab order. All I changed here is the accessible name: the previous native `title` became `aria-label` (to avoid a duplicate native tooltip next to the styled one), and `aria-label` on the entry requires a role, hence `role="button"`.

Making the palette actually keyboard-operable is out of scope here and a much bigger change — it needs the WAI-ARIA toolbar pattern (roving tabindex + arrow navigation + Enter/Space). A bare `tabindex="0"` wouldn't really solve it

assets/diagram-js.css
```diff
@@ -484,6 +491,49 @@
...
+
+.djs-palette-tooltip-shortcut {
+  padding: 1px 5px;
+
+  border-radius: 3px;
+
+  background-color: var(--palette-tooltip-shortcut-background-color);
+
+  font-family: monospace;
+  font-size: 11px;
```
barmac: can we relate this font size to anything? thinking about magic numbers

lib/features/palette/Palette.js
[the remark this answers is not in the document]
barmac: Agreed

lib/features/hover-tooltip/HoverTooltip.js
barmac, line 1: This is a feature, so let's place it in `features/hover-tooltip`.

lib/features/palette/PaletteTooltip.js
barmac: Sorry, I missed that message. 

> Different coordinate space, lifecycle, and trigger, so reusing tooltips would mean mixing two models into one service.

Sounds reasonable.

lib/features/hover-tooltip/HoverTooltip.js
barmac: However, I agree that `hover-tooltip` in general is an unfortunate name.

nikku (commented):
I second @barmac's question: There is three ways we can look at this contribution

(1) this is a general tooltip facility - and components provide details
(2) this is a general tooltip facility - palette provides titles - another component (?) injects keyboard shortcuts
(3) general tooltip facility + palette provides titles _and_ is keyboard aware

What is our stance? How do we envision the feature?

AlekseyManetov:
**On the facility itself** — it's already generic: [`hoverTooltip`](https://github.com/bpmn-io/diagram-js/blob/add-palette-tooltips/lib/ui/hover-tooltip/HoverTooltip.js#L111) renders opaque content and knows nothing about titles/shortcuts. Details always come from the consumer.

**On who provides the shortcut** — the palette knows nothing about entry actions or shortcuts, so they have to be provided to it. Two ways:

- **(3) — provide `shortcut` directly on the entry (current).** The bpmn-js provider fills it from the bindings map ([PaletteProvider.js#L123](https://github.com/bpmn-io/bpmn-js/blob/add-pallet-tooltips/lib/features/palette/PaletteProvider.js#L123), map at [BpmnKeyboardBindings.js#L35](https://github.com/bpmn-io/bpmn-js/blob/add-pallet-tooltips/lib/features/keyboard/BpmnKeyboardBindings.js#L35)). Already implemented in this PR - https://github.com/bpmn-io/bpmn-js/pull/2465
- **(2) — inject `shortcut` from a keyboard-aware provider** by redefining `getPaletteEntries` to augment entries and provide `shortcut` there. This inverts the dependency ("keyboard → palette" instead of "palette → keyboard"), but requires more work: entries have to expose their action + a `keyboard.getShortcut(action)` lookup.

Bottom line: either way the palette needs the `shortcut` on the entry to render it — the only question is how it's provided. (3) looks simpler, so I'd go with that unless we prefer to invest in (2). Or do you see any other options?

AlekseyManetov:
I've addressed the remaining review comments — moved the tooltip under `features/` and related the CSS font sizes to variables.

On the overall design I still think the current approach is the most reasonable (ref https://github.com/bpmn-io/diagram-js/pull/1082#issuecomment-5092908522) — happy to change direction if someone has a clear call for it.

Please take another look.

jarekdanielak (approved):
Looks good!

Consider renaming the feature to avoid the awkward "hoverTooltip", but at the end of the day it's just the name. Up to you.

PR #1090, opened 2026-07-28
chore: pin github actions to commit shas

Totara-thib:
### Proposed Changes

Small drive-by contribution: this pins the action references across the four workflows to their commit shas, versions kept as comments.

A tag like `@v4`, or the floating `@latest`, is a movable pointer: whoever controls the action, or anyone who compromises it, can re-point it and your next run executes their code with the job's token. Same pattern as the tj-actions/changed-files incident (CVE-2025-30066). With the sha, what runs is frozen to what you reviewed.

Notes:

- Renovate keeps working exactly as today: it understands sha pins with a version comment and will keep opening its bump PRs, same as its recent slack-github-action and setup-node updates.
- `bpmn-io/actions` is pinned to its current latest commit too. It is your own org so the risk model is different, drop that hunk if you prefer the moving tag.
- Heads up if the org uses an Actions allowlist: patterns written against tags (like `owner/action@v4`) stop matching once refs are shas and workflows refuse to start. Entries need to be `owner/action@*` in that case.

All shas were resolved from the upstream repos and cross checked against their tags. No logic changes, only the `uses:` lines. Found with the Plumber CLI (https://github.com/getplumber/plumber), verified on develop. I will also open a PR which adds it to CI so this does not quietly drift back, that one is a bonus, this PR stands on its own.

To try out: the diff is only `uses:` lines, and each sha can be checked against its version comment with `git ls-remote --tags https://github.com/OWNER/ACTION VERSION`.

### Checklist

  * [ ] __Link to related issue(s)__: none, drive-by security hardening
  * [ ] __Screenshots or short videos__: not applicable, CI-only change
  * [x] __Steps to try out__: the sha to version cross check above

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -14,16 +14,16 @@ jobs:
 
     steps:
     - name: Checkout
-      uses: actions/checkout@v7
+      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
     - name: Use Node.js
-      uses: actions/setup-node@v7
+      uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
       with:
         node-version: 24
         cache: 'npm'
     - name: Install dependencies
       run: npm ci
     - name: Setup project
-      uses: bpmn-io/actions/setup@latest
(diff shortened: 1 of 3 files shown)


nikku:
Thanks - we'll consider to do this at some point.

PR #1093, opened 2026-07-29
fix(outline): do not create outline for root element

### LINKED ISSUE #1092: Lazily-created outline for the diagram root element pollutes canvas.getActiveLayer().getBBox() / viewbox().inner
**Affected version:** diagram-js 15.23.0 (regression — not present in 15.18.1)

### Summary

Since `Outline` switched to lazy creation (`element.hover` / `selection.changed`, replacing the old `shape.added`/`connection.added` approach), it no longer excludes the diagram's root element. Hovering empty canvas fires `element.hover` for the root, and `Outline.prototype.createOutline` happily creates a `.djs-outline` `<rect>` for it — parented inside the root's own graphics group, which is never transformed (i.e. permanently sits at the origin).

That stray rect is real SVG geometry, so it's included when `getActiveLayer().getBBox()` computes the content bounding box (used by `canvas.viewbox().inner`). The reported min-x/min-y stays pinned near `(0,0)` forever, regardless of where the actual diagram content has moved.

### Why this matters

Any feature that trusts `canvas.viewbox().inner`/`getActiveLayer().getBBox()` to reflect real content bounds gets fed bad data. We hit this through `@bpmn-io/align-to-origin` (used by `camunda-dmn-js`'s DRD editor): its save-time `align()` reads that same bbox to decide whether the diagram needs realigning to the origin. Because the bbox never reflects the real content position, `align()` never converges — every single autosave recomputes a non-zero correction, executes `elements.alignToOrigin` (which moves all elements and compensates the viewport), which fires `commandStack.changed`, which triggers another autosave, forever.

Observable effect: a DMN diagram's autosave indicator gets permanently stuck on "Saving..." after appending a decision element on a freshly created diagram. Reproduces reliably; a previously-saved-and-reloaded diagram whose content already sits within align-to-origin's tolerance of the origin does not trigger it, which is why this looked at first like a test-flake rather than a library regression.

### Steps to reproduce

1. Create a new diagram in any diagram-js-based editor (bpmn-js/dmn-js) at diagram-js 15.19+.
2. Move the mouse over an empty area of the canvas (fires `element.hover` for the root element).
3. Inspect the DOM — a `.djs-outline` rect now exists inside the root layer's own graphics group (`transform: null`).
4. Call `canvas.getActiveLayer().getBBox()` (or `canvas.viewbox().inner`) — min-x/min-y is pinned near `-5` regardless of where actual elements sit.

### Root cause

In `lib/features/outline/Outline.js`, `createElementOutline` (invoked from the `element.hover` and `selection.changed` listeners) doesn't filter out root elements before calling `self.createOutline(element)`. In ≤15.18.1, `Outline` only listened to `shape.added`/`connection.added`/`shape.changed`/`connection.changed`, none of which ever fire for a root element, so this was never reachable. The new lazy-creation entry points changed that without adding an equivalent guard.

### Suggested fix

```js
import { isRoot } from '../../util/ModelUtil.js';

function createElementOutline(element) {
  if (isRoot(element)) {
    return;
  }
  self.createOutline(element);
}
```

(`isRoot` already exists in `lib/util/ModelUtil.js` and is used for equivalent guards elsewhere in the codebase.)

AlekseyManetov:
Closes https://github.com/bpmn-io/diagram-js/issues/1092

### Proposed Changes

**Root cause:** since outlines are created lazily (15.19+), hovering the root appends a `.djs-outline` rect to the untransformed root layer. That rect pollutes `getActiveLayer().getBBox()` / `viewbox().inner`, feeding wrong content bounds to consumers like align-to-origin.

**Fix:** skip outline creation for root elements (`isRoot` guard).

### Checklist

### DIFF
--- a/lib/features/outline/Outline.js
+++ b/lib/features/outline/Outline.js
@@ -18,6 +18,8 @@ import {
   isFunction
 } from 'min-dash';
 
+import { isRoot } from '../../util/ModelUtil.js';
+
 var DEFAULT_PRIORITY = 1000;
 
 /**
@@ -60,6 +62,13 @@ export default function Outline(eventBus, styles, elementRegistry) {
    * @param {Element} element
    */
   function createElementOutline(element) {
+
+    // root elements have no outline; creating one would append a never-updated
+    // rect to the untransformed root layer, polluting getActiveLayer().getBBox()
+    if (isRoot(element)) {
+      return;
+    }
+
     self.createOutline(element);
   }
 
--- a/test/spec/features/outline/OutlineSpec.js
+++ b/test/spec/features/outline/OutlineSpec.js
@@ -91,6 +91,40 @@ describe('features/outline - Outline', function() {
     ));
 
 
+    it('should not create outline for root on hover', inject(
+      function(canvas, eventBus, elementRegistry) {
+
+        // given
+        var root = canvas.getRootElement();
+
+        // when
+        eventBus.fire('element.hover', { element: root });
+
+        // then
+        var gfx = elementRegistry.getGraphics(root);
+
+        expect(domQuery('.djs-outline', gfx)).not.to.exist;
+      }
+    ));
+
+
+    it('should not create outline for root on selection', inject(
+      function(canvas, eventBus, elementRegistry) {
+
+        // given
+        var root = canvas.getRootElement();
+
+        // when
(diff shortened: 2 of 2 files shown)


nikku (approved):
Good catch, thanks ⭐

PR #1094, opened 2026-07-29
fix: support hiding element outlines

jarekdanielak:
### Proposed Changes

Adds a Canvas API to temporarily hide element outlines (toggles a `djs-outline-hidden` class on the canvas root). 

Lets consumers measure or serialize the diagram without lazily-created outlines skewing the bounds.

Related to https://github.com/bpmn-io/diagram-js/pull/1064

### Checklist

### DIFF
--- a/assets/diagram-js.css
+++ b/assets/diagram-js.css
@@ -153,6 +153,10 @@
   display: block;
 }
 
+.djs-outline-hidden .djs-outline {
+  display: none !important;
+}
+
 .djs-shape.connect-ok .djs-visual > :nth-child(1) {
   fill: var(--shape-connect-allowed-fill-color) !important;
 }


lib/core/Canvas.js
```diff
@@ -800,6 +808,20 @@ Canvas.prototype.toggleMarker = function(element, marker) {
...
+ * Hide all element outlines.
+ */
+Canvas.prototype.hideOutlines = function() {
+  svgClasses(this._svg).add(OUTLINE_HIDDEN_CLS);
+};
+
+/**
+ * Show all element outlines.
+ */
+Canvas.prototype.showOutlines = function() {
```
nikku: I'm not sure we need the APIs - outline is not core to the canvas, it does not _know_ about it. Or do I see that wrong? If at all, then outline feature should own that API.

I'd also be happy to just ship the CSS class - so with internal knowledge ("for now") the outline can be hidden.

jarekdanielak: Let's ship just the class.

PR #1105, opened 2026-09-23
fix(util): prevent endless loop when laying out text

kern0x1b:
### Proposed Changes

`layoutNext` checks whether a line fits against `Math.round(maxWidth)`, but `shortenLine` shortens it against the unrounded `maxWidth`. When the measured width of a line lies between the two, `shortenLine` computes a target length longer than the line, `semanticShorten` returns the whole line, and the loop never ends. The browser tab freezes.

A real diagram hits it: an end event labelled `Fall erfolgreich bearbeitet`, measured at 124.12px in 11px Arial, with a label box of 124.457px from the DI. `Math.round(124.457)` is 124, the line is not below it, and shortening it by the ratio `124.457 / 124.12` gives back the same line. Importing the diagram into `bpmn-js` 18.29.1 hangs.

This shortens against `Math.min(maxWidth, textBBox.width)`. The loop only continues when the line did not fit, so the ratio passed to `shortenLine` is at most 1 and every step makes the line strictly shorter; the loop ends after at most one step per character. Where the line is wider than `maxWidth`, which is every case that terminated before, the value is `maxWidth` as before, so no existing layout changes.

To check that last claim, I ran the old and the new `layoutNext` against each other on 200,000 random lines, character widths and box widths, half of them fractional: the old one hangs in 1,110 of them, the new one in none, and in every case where the old one terminates both produce identical lines.

The added test picks a label whose width has a fractional part below 0.4 and lays it out in a box `floor(width) + 0.49` wide. Without the change, the test run disconnects with a ping timeout because the tab hangs. With it, the label breaks.

### Checklist

- [x] Covered by a test
- [x] `npm run all` passes locally (2201 passed; the occasional failure in the snapping, resize and popup menu specs also happens on `develop` without this change)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lib/util/Text.js
```diff
@@ -222,7 +222,7 @@ function layoutNext(lines, maxWidth, style) {
       return fit(lines, fitLine, originalLine, textBBox);
     }
 
-    fitLine = shortenLine(fitLine, textBBox.width, maxWidth);
+    fitLine = shortenLine(fitLine, textBBox.width, Math.min(maxWidth, textBBox.width));
```
nikku: The computation should move into `shortenLine`?

nikku:
One edge where the clamp is not fully behavior-preserving: in 2/200k differential fuzz samples it splits a trailing soft hyphen onto its own line — at ratio exactly 1.0 (`target == width`), the strict `<` boundary in `semanticShorten` drops a trailing `­` that the unclamped path translates to `-`. Rare, and it still terminates, but worth knowing.

For reference, we sketched an alternative that drops the rounding in the fit guard instead (`width <= maxWidth`), which also guarantees labels never overflow their box: #1106.

kern0x1b:
Thanks - #1106 fixes the cause rather than the symptom: the fit check and the shortening now use the same width, a rejected line is always strictly wider than the box so every step makes progress, and labels no longer overflow by a fraction of a pixel. Happy to close this one in favour of it.

nikku:
Thanks again for the report and the initial fix @kern0x1b :star: 

Released via [`v15.27.1`](https://github.com/bpmn-io/diagram-js/releases/tag/v15.27.1)

